PR #15, opened 2022-03-07
Update faq.md

Vaibhavs10:


data/faq.md
```diff
@@ -2,9 +2,9 @@
 
 ## **Q. When and where is EuroPython 2022 taking place?**
 
-A. After 2 years of being fully remote, EuroPython 2022 will be held on**11th-17th July**and it will be both in person and virtual.
+A. After 2 years of being fully remote, EuroPython 2022 will be held on **11th-17th July** and it will be both in person and virtual.
 
-The in-person conference will be held at [**The** **Convention Centre Dublin (The CCD**)](https://www.theccd.ie/)** in Dublin**,**Ireland**.
+The in-person conference will be held at [**The** **Convention Centre Dublin (The CCD**)](https://www.theccd.ie/) **in Dublin,** **Ireland**.
```
artcz, line 7: I think this could be simplified to just using `**(All the content)**` instead of splitting it to multiple parts

Vaibhavs10: good catch, fixed it..

PR #27, opened 2022-03-17
put buy-ticket button up top too

hypha:
put buy-ticket button up top too just in case people know the drill and want to go to the shop directly.


patrick91:
Done in 8a0ef76 :)

PR #33, opened 2022-03-26
Initial sponsor page implementation

patrick91:


hypha:
Missing note.tsx is probably why it fails. Is it possible to find a nice components to list a few add-ons?

PR #38, opened 2022-04-08
sponsor_logos

hypha:
1.Put all tiers in order in sponsor.json, with some examples. 
2. Organised all sponsor logos in img/logos/sponsor_logos/ so they are in one place. 
3. Deleted the example sponsor logos in img/ and img/logos. 
@patrick91 Is it OK with 2+3?


patrick91 (approved):
Awesome, thanks for cleaning up the image folder :D

PR #40, opened 2022-04-14
social-cards-updates

hypha:


patrick91 (approved):
very nice!

hypha:
- I added cards to voting and tickets. 
- I standardised the social cards a little bit: logo, page, two lines of blurb. All cards are of size 1686 x 888. 
- I changed the font to the same as the logotype's. 
- Also uploaded a template in svg so everyone can quickly create a new one.

PR #44, opened 2022-04-25
Closing the voting

artcz:


hypha (approved):
I've also put the finaid CTA back, rearranged the header and updated the FAQ milestones.

PR #46, opened 2022-04-28
collapsible sponsor info page

hypha:
@patrick91 PR made. As mentioned, the h2 is in a new line and the position is not quite right. Not worth spending much time on it until we know if we like this way of displaying the page.


Ladybiss (approved):
LGTM!
We can probably remove that booth_silver_mock.jpg and the directory since we are not using those anymore.

Thank you so much Raquel 🌻

PR #47, opened 2022-04-29
Add Ebury logo

Ladybiss:


Ladybiss:
not sure what I am missing but the logo is not visible in the preview 🤔

hypha (commented):
Just resaved it in inkscape. I often encounter adobe ghosts :P
I think it looks good now?

patrick91 (approved):

PR #48, opened 2022-04-29
1pass logo

Ladybiss:


artcz:
image
I think something went wrong here :sweat_smile:

Ladybiss:
> image I think something went wrong here 😅

yep 😁 That's one of the reasons I didn't merge yet!

hypha:
> > image I think something went wrong here sweat_smile
> 
> yep grin That's one of the reasons I didn't merge yet!

should look good now?

patrick91:
@hypha @Ladybiss I've done some changes to the css 😊 they might break when we get more lines of sponsors but I'll happily fix it then :)

hypha:
> @hypha @Ladybiss I've done some changes to the css 😊 they might break when we get more lines of sponsors but I'll happily fix it then :)

Your promise is better than anything we can hope for @patrick91 :)

PR #56, opened 2022-05-06
Add initial schedule

patrick91:


hypha:
>     are we going to support talks spanning multiple time slots? see talk called "A talk spanning two time slots sloth"
Yes. But ideally, a 45 min talk in one track will have 15+30 min ones in the same slot in other tracks, so there is no gap. But if we have a gap by having breaks or something else, e.g. track 1: 45 min; track 2: 15 min + 5 min break + 30 min , then maybe we can fill it with some nice image, like shamrock or EP logo? Would any of that be possible?
 
>   If we have talks spanning multiple time slots, do we show them in all the time slots they overlap with? for example, for the talk above, we'd see it in both the 14:10 time slot and the 14:35. I'm asking this because it might be useful to know, or maybe on mobile we also the durations of the talks?

In the browser version, I think it makes sense to show the block by the longest talk in the slot, as discussed in point 1. Don't have good ideas about mobile, but +1 on showing durations! 
 
>      should we add some filters at the top? we could allow hiding tutorials (if we have them in parallel) so we get more real estate. If we don't have tracks separated by topics maybe more filters won't make sense

Hmmm tutorials take place only on Mon and Tue, right? So what would you be hiding? If you filter by level for example, wouldn't you see mostly gaps? Filters certainly make sense on the session list? 

>     * [ ]  Will we have taglines for speakers? the design has them :)
I think our design review says we won't have taglines for speakers, but only for keynotes. We did say we might want to replace tagline with their company name though. 
>     * [ ]  Do we have talks with multiple speakers? in that case do we show taglines?
Yes, we do.

patrick91:
I've done quite a bit of progress, we also have individual pages for days (and for talks) and a sessions lists page, here's some examples:

1. schedule: https://ep2022-git-feature-schedule-europython-society.vercel.app/schedule/2019-07-10
2. single talk: https://ep2022-git-feature-schedule-europython-society.vercel.app/talks/analytics-with-%F0%9F%90%BC-pandas-and-jupyterlab
3. sessions: https://ep2022-git-feature-schedule-europython-society.vercel.app/sessions


they both need to get real data. Once we have a real list of sessions I'll update the code to render the right data 😊

so the way I setup the schedule is like this:

```js
{
  "days": {"2019-10-10": {...}}
}
```

so that each day will have its own set of rooms and talks, hopefully that's ok.
You can look at the json for all the data I used, the JSON is not final yet, I want to remove things like ev_custom, tt_duration and so on (there's no need for that in the json)

Also the current session list is coming from the schedule data, which is not ideal since we'll have the list of talks first and then the schedule, I can update that. I think we can do something like this:

1. `all_session.json` a JSON with all the approved sessions and information, maybe keyed by id
2. `schedule.json` a JSON with the schedule, similar to what we have, but it links to `all_session.json` for speaker info, titles and so on


or we can use a sqlite db, that works fine too 😊

PR #63, opened 2022-05-13
Add Optum logo.

Ladybiss:


Ladybiss:
poking a couple of people here @patrick91 @artcz because after adding the new sponsor Optum I see a result that's maybe a bit ugly: 

image

I think that the logo is too close to the green line under, WDYT?

patrick91:
@Ladybiss fixed!

PR #70, opened 2022-05-23
keynoter/laura

hypha:


patrick91:
@hypha I've rebased this to include the automatic social card 😊 should work once merged (the preview won't work because it is hardcoded with ep2022.europython.eu) :)

PR #89, opened 2022-06-07
Add initial schedule from api

patrick91:


Vaibhavs10:
Hey Patrick!

This is lovely! Couple of things that we should ideally fix:
1. The tutorials on Monday and Tuesday only show for 90 minutes: https://ep2022-git-feature-schedule-from-api-europython-society.vercel.app/schedule/2022-07-11

2. We ideally need a break in the middle so we need to figure out a way to break each tutorial into two parts. The flow is something like this:

> first tutorial starts at 09:30, we break at 11
> then tutorial gets over at 12:30

> lunch break 12:30 to 13:30

> second tutorial starts at 13:45, we break at 15:15
> then tutorial gets over at 16:45

3. Some of the talks and tutorials don't have talk/ tutorial page not linked to them. for ex: https://ep2022-git-feature-schedule-from-api-europython-society.vercel.app/schedule/2022-07-12 does not have any of the tutorials page linked to the description.

4. Some of the talk durations are a bit longer than what they actually should be: https://ep2022-git-feature-schedule-from-api-europython-society.vercel.app/schedule/2022-07-14 ex: `Online voting system used for primary elections for the French Presidential, must be secure right?` this is a 45 minute long session and it appears to be much more than that.

5. Similarly with posters, some of the posters have the wrong duration ex: `EModelRunner: a Python package to run online available biological neuron model implementations` on https://ep2022-git-feature-schedule-from-api-europython-society.vercel.app/schedule/2022-07-14

6. Just from a design standpoint, is it possible to have the speaker names and the horizontal line to the bottom right? just so they don't look all over the place  in the schedule

Thanks again for putting this together! <3

artcz:
Just making a note here before I forget: for the tutorial split in half, we can either 

a. change them to be 90 minutes in pretalx and then using the slots feature schedule them twice (just like sponsor talk is scheduled 6 times right now); then in the API we can change duration to every session that's type tutorial to be 180; and return it twice in the API just like in the schedule editor. - this has added benefit so that we can add a proper break in pretalx too.

or

b. keep it as-is in pretalx, but do all the post processing and splitting directly in the API (so that later the website only needs to place them on the grid, and doesn't have to bother with all the logic. In this case it would be taking a 180 minute session, creating a duplicate of it, changing both to 90 minutes; then moving the second session to be 15 minutes after first session ends.


Either one would result in the API already returning two sessions per tutorial that have 90 minute duration and correct timestamp; we just need to decide which one do we prefer.
Both seem to be similar amount of work required; (a) has added benefit of having same layout inside pretalx as well. (b) has an added benefit that we don't have to deal with multiple slots.

Vaibhavs10:
I'd honestly say we go #2, I don't want tutorials speaker to get confused by having multiple sessions scheduled under their name.

PR #93, opened 2022-06-08
fix typo in timeline

yednapg:
Fixed Typo in Timeline.
image


hypha:
> Fixed Typo in Timeline. image

Thank you for the fix!

PR #121, opened 2022-06-23
Update mentored-sprint.md

Cheukting:


hypha:
Maybe still mention Open Space on top of the room name?

PR #147, opened 2022-07-04
Add Virtual Swag Page

hypha:


hypha:
@patrick91 I have some questions:
1. Currently, if I click the PDF, it downloads. How do I open it in another browser?
2. Could I make the img a bit larger?

patrick91:
> Currently, if I click the PDF, it downloads. How do I open it in another browser?

this works fine for me, what browsers are you using?

> Could I make the img a bit larger?

Yes, but we might want to use a higher quality image

hypha:
> this works fine for me, what browsers are you using?
Firefox

> Yes, but we might want to use a higher quality image
pushed for a higher-res now.

PR #157, opened 2022-07-08
Added Mu to sprints.

ntoll:
What it says on the tin.


data/pages-content/sprints.md
```diff
@@ -36,3 +36,10 @@ Many thanks !
 - Contact: helpdesk@europython 
 - Links: https://your_project_url/README.src
 
+### Mu Editor 🐮 🐍 🎉
+
+- Number of people: 4-5 core maintainers + anyone who wants to join us.
+- Room: The Forum (otherwise somewhere clearly signposting us)
```
gilgamezh, line 42: We have to fix the rooms. (it's in a different PR) maybe just use: Wicklow Hall 2A/B

PR #212, opened 2022-11-05
Add case for missing http in speaker URLs

### LINKED ISSUE #205: Speaker homepage URL
If a speaker's homepage URL (on the individual speaker page) does not start with http(s)://, it appends https://ep2022.europython.eu/speaker/ before the URL.

Sangarshanan:
should fix #205


hypha:
LGTM! Thanks!

PR #216, opened 2022-12-09
Add color tweaks

patrick91:


hypha:
Funnily, none of the content on the collapsible header on the second row are clickable. You move to click and it disappears.

patrick91:
Add more colors and fixed the bug with the header :)

PR #217, opened 2022-12-21
Add burst image in vector format

hypha:


hypha:
@patrick91 I uploaded the svg version so we can make references to all the svg files in one place for design purposes.

PR #220, opened 2023-02-01
Add clacks of remembrance to our base template

ntoll:
Context: In the Terry Pratchett novel "Going Postal", a telegraph-style system known as "Clacks" was used to pass the name of a deceased character endlessly back and forth, keeping their memory alive. But where the book had "GNU John Dearheart" -- the prefix being a basic code to instruct clacksmen to pass on, not file, and return the message -- we add meta headers to our base template in a silent but appropriately geeky tribute and act of remembrance to those in our EuroPython family we have lost.

### DIFF
--- a/components/clacks.tsx
+++ b/components/clacks.tsx
@@ -1,0 +1,19 @@
+// https://xclacksoverhead.org/home/about
+// we're using `dangerouslySetInnerHTML` here because nor Next.js nor React support
+// html comments in JSX, and we want to keep the In remembrance section in the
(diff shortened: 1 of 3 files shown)


patrick91:
@ntoll thanks for doing this! I've updated the code with a workaround to make print the HTML comment in the output 😊

ntoll:
@patrick91 thank you so much for doing this. It means a lot. Here's hoping we don't have to add to it any time soon.

PR #230, opened 2023-02-20
Header, Footer & Home

hypha:
This will get merged last, after all other content page goes live.


hypha:
@patrick91 I'll leave the homepage components to you! Feel free to leave content updates to me.

Issue #237, opened 2023-02-26
Anchor links for headers don't jump to the relevant header


SebastiaanZ:
If you add a markdown header to a "pages-content" file, the Next.JS rendering will automatically wrap the header in a link with an anchor for that header (e.g., [https://ep2023.europython.eu/mentorship#mentee][mentee-anchor] for the header "Mentee" on the mentorship page). The problem is that the target anchor name isn't added as the `id` attribute of the header HTML tag.

This means that anchor links don't actually jump to the relevant location on the page.

[mentee-anchor]: https://ep2023.europython.eu/mentorship#2-workshops

SebastiaanZ:
The same goes for a markdown footnote. The footnote will be rendered correctly, with an anchor link at the place of the footnote reference to the footnote, but the footnote itself (rendered at the bottom) will not have the necessary `id` to make the footnote link work.

SebastiaanZ:
> @SebastiaanZ I fixed the headings :)
> 
> do know where we are using footnotes? I'll fix them as well :)

@patrick91 There's one on the new finaid page, `/finaid`.

patrick91:
@SebastiaanZ fixed in ae99a41! thanks 😊

PR #238, opened 2023-02-26
Update finaid page with information for EP 2023

SebastiaanZ:
This commit contains an updated finaid page with information relevant to EuroPython 2023, including a link to the form.

To highlight the two separate rounds of finaid, I've also added two SVG files that provide a nice accent to the relevant lines in the timeline:

image

Note that the footnote for the deadline of round 2 does not actually work properly at this moment. This is due to a missing anchor or id in the rendering of the actual footnote at the bottom of the page. This problem is probably related to the lack of anchor ids on header tags, as described in #237.


SebastiaanZ:
I noticed that I did not add a link to the finaid page to the navigation menu yet. Where should it go?

PR #241, opened 2023-02-28
Add ticket page, pre-sale & FAQ

hypha:
- [x] Undecided information, including time and price of the sales, are left with #FIXME
- [x] clean up FAQ
- [x] add ticket to menu

PR #242, opened 2023-02-28
Add non-clickable button component

hypha:
This is a non-clickable button - to denote that when we are ready, it will be a clickable link, changing to `<ButtonLink>`


hypha:
@patrick91 no idea if it makes sense... I wanted to use that in [here](https://github.com/EuroPython/website/pull/241/files#diff-85425451be46ea2104b2e34f1abada058507e784ba007f2c3309dd745a319c7eR54), so made something.

PR #251, opened 2023-03-14
Remove extra Net price where it doesn't apply

artcz:


hypha (approved):
Consistency thanks you, @artcz

PR #291, opened 2023-05-12
Add jetbrains to sponsors.json

hypha:
@patrick91 

- [x] seeing that we have 4 platinum max, the last one now gets into the second line. Can we make each platinum logo bigger, occupying 2x2 instead? 
- [x] I am also messing up with the types of [components/sponsor-tiers/sponsor-tiers.tsx](https://github.com/EuroPython/website/blob/dedaee73d924febda5fdcf7f9be142d5d7433fc0/components/sponsor-tiers/sponsor-tiers.tsx#L153)


patrick91:
all done!

PR #301, opened 2023-05-17
Enable static params

patrick91:


patrick91:
this takes too long, I might have done something wrong, let's not merge it for now :)

PR #322, opened 2023-06-07
Add childcare clarification

hypha:


hypha:
@thepetk I added some clarification for the childcare bits: whether they need another ticket and whether teens can use childcare. Could you check if it looks OK?

thepetk (approved):
/lgtm based on the overall support tickets we had so far. Suggested, not requested, change could be to mention support@europython.eu as parents might have more questions for this subject.

PR #336, opened 2023-06-16
add parcel info for exhibit

hypha:


hypha:
@Kislovskiy could you check if the parcel information added makes sense? I am not so share about how to best list it in the due date section at the top.

data/pages-content/sponsor/information.md
```diff
@@ -203,32 +206,27 @@ distribute them at your booth. We strongly encourage you to distribute gifts
...
+- Martin ULIČNÝ
+  - Tel: +420 242 405 160
+  - E-mail: martin.ulicny@dbschenker.com
 
 Here are shipping information provided by DB Schenker:
 
-- [Shipping Guidelines](TBD).
-  **Important: Please provide DB Schenker your Company name, point of contact name
-  and email.** Contact DB Schenker for Details.
+- [Shipping Guidelines & Address (.docx)](https://docs.google.com/document/d/1tgDfBikrFo1e9rtWWCNgNbe8q2I8mELl/edit).
```
Kislovskiy, line 220: Maybe we should "Copy link" instead, so that people can view the document, but not edit?

data/pages-content/sponsor/information.md
```diff
@@ -203,32 +206,27 @@ distribute them at your booth. We strongly encourage you to distribute gifts
...
+  - **This is the [sample label](https://docs.google.com/document/d/1AnNPyK7D9cnEIhMXsAU2VK5-3mxaLEzs/edit) provided. Please contact DB Schenker for Details.**
 
-- [Freight Order link](TBD).
-- [Delivery address and Shipping Label](TBD).
+- [Shipping Quotes (.docx)](https://docs.google.com/document/d/1shj6VhH4Wu4y76m-b5AtLviYbQsVGJY5/edit).
```
Kislovskiy, line 223: The same here, I think the view role would work better.

data/pages-content/sponsor/information.md
```diff
@@ -203,32 +206,27 @@ distribute them at your booth. We strongly encourage you to distribute gifts
...
+  - **Contact DB Schenker directly to discuss pricing and conditions**
 
-**Deadlines:**
+**Deadlines**
+- Deadlines are specified in [Shipping Guidelines & Address (.docx)](https://docs.google.com/document/d/1tgDfBikrFo1e9rtWWCNgNbe8q2I8mELl/edit).
```
Kislovskiy, line 227: And here 😀 edit -> view

data/pages-content/sponsor/information.md
```diff
@@ -10,6 +10,9 @@ subtitle:
 
 
 ### Due Dates of Deliverables
+<h5> Variable Dates</h5>
```
Kislovskiy, line 13: To be honest, I had to google what does this phrase mean 🙃
Maybe we could use something like "Flexible Dates" instead?

data/pages-content/sponsor/information.md
```diff
@@ -10,6 +10,9 @@ subtitle:
...
+- [Parcel Delivery for Booth Items](#parcel): (all exhibitors wishing to ship exhibit items)
```
Kislovskiy, line 14: Maybe "wishing" -> "who want", it would look like this: `(all exhibitors who want to ship their exhibit items)`

Kislovskiy: I would use the same title here and on line 199. It's a bit confusing to click on "Parcel Delivery for Booth Items" and after click on "Parcel Delivery & Gifts" to see the details.

Kislovskiy (commented):
Thank you for putting the shipping information to the website 😎
I have a few suggestions

data/pages-content/sponsor/information.md
hypha: yup, the doc's permission was set to view only but /view would be more accurate indeed!

data/pages-content/sponsor/information.md
hypha: I've now changed it to Parcel Delivery & Swag for both.

data/pages-content/sponsor/information.md
hypha: I don't want to give the impression that the dates are flexible. So I now just put it as "go read for details".

Kislovskiy: sounds good

Kislovskiy (approved):
Thank you for doing this 😊 Looks good to me

PR #366, opened 2023-07-12
add pyladies lunch

hypha:


hypha:
@whykay could you sanity check my insanity? <3

whykay:
Updating here as well to say it all looks good. You can do a shout out about allies coming along during conference week, right now we want to make sure PyLadies community knows about this first.

PR #399, opened 2023-09-20
List ep2023 cut videos

hypha:


app/(pages)/live/[room]/page.tsx
```diff
@@ -50,9 +50,15 @@ export default function LivePage({ params }: { params: { room: string } }) {
 
       <div className="h-12" />
 
+      <Title>
+        <Link href="https://www.youtube.com/playlist?list=PL8uoeex94UhFcwvAfWHybD7SfNgIUBRo-">
+          Edited video playlist 📼
```
Comment (hypha) [inline app/(pages)/live/[room]/page.tsx:55]: @patrick91 is it a good idea to list the edited playlist here as well? You got better ideas?

PR #414, opened 2023-12-13
Don't deploy to Chromatic from dependabot branches

Kislovskiy:


Kislovskiy:
@patrick91 do you think it's okay to skip Chromatic deployment from dependabot branches?

patrick91 (approved):

PR #449, opened 2024-02-04
Fix mentorship/programme registration link

cmaureir:


cmaureir:
On the website, we have no need to have the outdated link with the redirection, we can use the correct url.

artcz (approved):

artcz (approved):
Thanks!

PR #481, opened 2024-03-02
Fix CFP/voting typo

artcz:


artcz:
Thanks @baradrb for reporting :)

PR #489, opened 2024-03-08
revive header logo to return home

hypha:
It wasn't easy to go back to home page otherwise :)

### DIFF
--- a/components/header/header.tsx
+++ b/components/header/header.tsx
@@ -96,7 +96,7 @@ const HeaderActions = ({ mobile = false }: { mobile?: boolean }) => {
   );
 };
 
-export const Header = ({ inverted = false }: { inverted?: boolean }) => (
+export const Header = ({ inverted = true }: { inverted?: boolean }) => (
   <header className={clsx("p-6 flex items-center relative z-40", {})}>
     <input
       type="checkbox"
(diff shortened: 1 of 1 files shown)


patrick91:
Nice! do you think we should change the logo to have better contrast with the bg? 

CleanShot 2024-03-08 at 12 53 33@2x

hypha:
> Nice! do you think we should change the logo to have better contrast with the bg?
> 
> CleanShot 2024-03-08 at 12 53 33@2x

I was thinking the same thing and thought the pink & blue one might be better for contrast against white. 
image

thepetk (approved):
lgtm

hypha:
@patrick91 @thepetk @artcz perhaps the current blue and pink looks better for contrast ( https://github.com/EuroPython/website/pull/489/commits/eac16dc80becf28620dd97891f15f31e25c1ad4b) like this?
image
 

I think this one is part of the hero and not componentised separately as a logo, which might not be easy to use in comparison? 
image

If you are happy for the current version, feel free to merge or discuss which colours you'd like or take over entirely ofc :)

hypha:
hey hey! any thoughts on whether the current version can be merged or it should be modified in any specific direction? :hugs:

PR #526, opened 2024-03-20
fix numbers and update homepage deadline

artcz:


artcz:
image


I just updated it now so we have link on the homepage. Feel free to change the text or the picture :)

cmaureir (approved):

PR #528, opened 2024-03-21
Mobile font fixes

### LINKED ISSUE #527: Texts do not fit the screen on mobile
Here are two examples:

I'm using Chrome in Android.

I'm not sure if these kind of issues are a priority or not, so just opening one just in case :)

patrick91:
Closes #527

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -27,9 +27,9 @@ export default defineConfig({
       [
         rehypeAutolinkHeadings,
         {
-          behavior: "prepend",
+          behavior: "append",
           content() {
-            return [h("span.heading-link", "#"), h("span.icon.icon-link")];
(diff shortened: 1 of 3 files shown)


astro.config.mjs
```diff
@@ -27,9 +27,9 @@ export default defineConfig({
       [
         rehypeAutolinkHeadings,
         {
-          behavior: "prepend",
+          behavior: "append",
           content() {
-            return [h("span.heading-link", "#"), h("span.icon.icon-link")];
+            return [h("span.heading-link", "#")];
```
patrick91, line 32: I moved the # on the right on mobile, so it is always visible: 
CleanShot 2024-03-21 at 10 56 38@2x

PR #533, opened 2024-03-23
update sponsor information page

aramboi:


artcz (approved):
LGTM! I think the deadlines look good enough to merge now; in case we have some updates from EXPO we can adjust the content later – most of the deadlines are in June/July anyway. The April deadline for 10% discount is already fixed.

PR #547, opened 2024-03-28
Add pre-commit checks and fix its complaints

### LINKED ISSUE #497: Investigate if we can add precommit hook for commiting large files to the repository accidentally
This is mostly to check if we don't add big binary files or images. 

We can also separately check this on CI.

egeakman:
Closes #497 

We can change the ``maxkb`` argument to our needs.

Added some handy checks as well.

Since we are using [``pre-commit``](https://pypi.org/p/pre-commit), we might want to add that to a ``requirements-dev.txt``.


patrick91:
@egeakman this is amazing, can we add some instructions to the readme for installing pre-commit?

Also, maybe it is worth adding https://pre-commit.ci/ to this repo, so we are sure we always run it 😊

we should also probably add prettier!

egeakman:
I think it's a good idea to add pre-commit.ci here. I can't add it though, maybe a repository admin can?

egeakman:
Prettier's result is pretty noisy, it changes ~43 files. I'll push the changes, but we probably should check the preview carefully.

patrick91:
@artcz would you be ok with pre-commit ci? we could do a github action, but pre-commit ci is faster and free 😊

egeakman:
Prettier somehow breaks it.

It turns;

``{/*`` into ``{/\*``

and

``*/}`` into ``\*/}``

egeakman:
Found lots of similar issues like [this](https://github.com/prettier/prettier/issues/15163) and [this](https://github.com/prettier/prettier/issues/12209)...

Turns out that ``prettier`` only supports MDX v1.

We can add an ignore flag to each problematic line, or we can simply skip all MDX files.

egeakman:
@patrick91 hi! I reduced the amount of changed/affected files by ~half.

``prettier`` now:
- doesn't touch the lock file,
- doesn't add a trailing comma everywhere,
- keeps the tab width consistently at 2 spaces

How should we proceed?

patrick91 (approved):
Awesome, thanks!

PR #549, opened 2024-04-01
Update website content

thepetk:
Updates programme content, removes the community voting deadline and adds a sponsor one


cmaureir (approved):

src/content/deadlines/00_cfp.mdx
```diff
@@ -1,10 +0,0 @@
----
```
patrick91, line 1: did you delete this because it shows on the home page? 😊

thepetk: yeap I'd like to remove that in favor of sponsors. Preview was ok but let me know if I have to update anything else!

patrick91: I've done that in code, so we can keep the deadlines, and only show the last 3 😊

thepetk:
Done!

PR #557, opened 2024-04-07
Fix sponsor page for mobile view

### LINKED ISSUE #534: Sponsor tier layout
- On mobile (1080 x 2400) and laptop full HD, the ribbons are overshadowing "available" in the last four packages. 
image

- Platinum "X" is overlapping with the ribbon
image

egeakman:
- Fix the ribbon covering the text (a better looking solution would be nice instead of this fix)
- Fix buttons sticking too close

image -> image

image -> image

closes #534

### DIFF
--- a/src/components/sponsor-tiers/sponsor-tiers.tsx
+++ b/src/components/sponsor-tiers/sponsor-tiers.tsx
@@ -72,7 +72,10 @@ const SponsorTier = ({
       : price;
 
   return (
-    <div className="bg-white text-black rounded-2xl p-6 relative not-prose">
+    <div
+      className="bg-white text-black rounded-2xl p-6 relative not-prose"
+      style={{ zIndex: 0 }}
+    >
       <div className="h-[160px]">
         <Ribbon
           className={clsx("absolute right-6 -top-6", {
@@ -84,6 +87,7 @@ const SponsorTier = ({
             "text-sponsor-bronze": title === "Bronze",
             "text-sponsor-patron": title === "Patron",
           })}
+          style={{ zIndex: -1 }}
         />
 
         <Title level={3} className="mt-0 !mb-2">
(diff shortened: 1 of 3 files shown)


hypha:
closes https://github.com/EuroPython/website/issues/534

src/components/sponsor-tiers/sponsor-tiers.tsx
```diff
@@ -72,7 +72,10 @@ const SponsorTier = ({
       : price;
 
   return (
-    <div className="bg-white text-black rounded-2xl p-6 relative not-prose">
+    <div
+      className="bg-white text-black rounded-2xl p-6 relative not-prose"
+      style={{ zIndex: 0 }}
```
patrick91, line 77: I think you can do `z-0` with tailwind, but what does it do in this case?

src/content/pages/sponsor.mdx
```diff
@@ -41,7 +41,7 @@ accessible to most organisations keen to help the community.
 <div style={{textAlign: "left", marginBottom: "30px"}}>
```
patrick91, line 41: could be nicer to use tailwind's space-y class here, instead of using a class on the button: https://tailwindcss.com/docs/space

what do you think? 😊

patrick91 (changes requested):
left some comments! the output looks great, but I think we can improve the code a bit 😊

src/components/sponsor-tiers/sponsor-tiers.tsx
egeakman: Yep, z-0 is nicer. Without setting the z-index, the ribbon is sent all the way back:
image

And if the ribbon's z-index is anything other than ``-1``, it stays on top of the text.

patrick91 (approved):

PR #571, opened 2024-04-11
update sponsor booth customization info

aramboi:


Kislovskiy (approved):
Looks good, thanks @aramboi

PR #601, opened 2024-05-13
Download sessions and speaker data

patrick91:


.pre-commit-config.yaml
```diff
@@ -17,5 +17,5 @@ repos:
     hooks:
       - id: prettier
         args: ["--tab-width", "2", "--trailing-comma", "es5"]
-        exclude_types: ["mdx"]
+        additional_dependencies: ["prettier", "prettier-plugin-astro"]
```
patrick91, line 20: took the opportunity to fix precommit for astro 😊

scripts/download-data.py
```diff
@@ -0,0 +1,70 @@
...
+
+    output_dir.mkdir(parents=True, exist_ok=True)
+
+    for key, value in data.items():
+        filename = f"{key}.mdx"
+        path = output_dir / filename
+
+        content = value.pop(content_key) or ""
+
+        content = content.replace("<3", "❤️")
```
patrick91, line 34: There's a <3 that was breaking the parsing, so I fixed it like this for now 😊

PR #617, opened 2024-05-27
Add volunteering card

thepetk:
As the finaid deadline has passed, we can add a card for the new page regarding onsite volunteers.

Added [DO NOT MERGE] label as I think it needs to be approved from ops <3


egeakman (approved):
LGTM! The image is a little big as a file, but I think it gets optimized before deployment, so no issues on the user-side

thepetk:
@egeakman I've gone through again and found a minor text fix I needed to do. Could you review once more?

egeakman (approved):

PR #622, opened 2024-05-29
Fix sourceless image tags

egeakman:
Here is an example:

image

and

image


egeakman:
Or we can assign a default image over at [programapi](https://github.com/EuroPython/programapi)

cmaureir (approved):
the change makes sense to me. Thanks @egeakman

patrick91 (approved):

PR #640, opened 2024-06-07
Reuse filtering/listing code in sessions/talks/posters/tutorials

egeakman:


src/components/sessions/list-sessions.astro
```diff
@@ -0,0 +1,68 @@
+---
+const { sessions } = Astro.props;
+import Prose from "../prose/prose.astro";
+import { Tag } from "../tag/tag";
+import SessionSpeakers from "../session-speakers.astro";
+
+type Session = {
+  data: {
+    title: string;
+    // speakers: ???;
```
egeakman, line 10: @patrick91 How should we type it?

PR #646, opened 2024-06-10
schedule

patrick91:
Rebuild of the schedule to be more flexible, should hopefully accomodate much more use cases


cmaureir:
Not sure if there is an easy way of adding `Title (Speakers)` for the Lunch-poster session within the Lunch square would be a good idea...or if it will be more confusing to people. Initially I was thinking of "Lunch + Poster Sessions" kind of text for the lunch break, then each title could be clickable to get to the page with more info.

PR #670, opened 2024-06-12
Insert links to tags, add some style :), fix mobile view for tags

### LINKED ISSUE #634: Insert links to session tags
image

Insert links to tags.

For example, clicking on the first tag (session type) would lead to sessions page with session type filtered

Second -> sessions page with track filtered

Third -> sessions page with level filtered

egeakman:
Closes #634

The changes on filter.astro is to be able to accept filters from URL parameters.

The changes on tag.tsx is to be able to pass a link.

Fancy hover style: (it turns blue)
image

Better mobile view:
image

### DIFF
--- a/src/components/tag/tag.tsx
+++ b/src/components/tag/tag.tsx
@@ -1,30 +1,33 @@
 import clsx from "clsx";
 
-export const Tag = ({
-  children,
-  className,
-}: {
+interface TagProps {
   children: React.ReactNode;
   className?: string;
-}) => {
+  href?: string;
+}
+
+export const Tag = ({ children, className, href }: TagProps) => {
+  const TagElement = href ? "a" : "span";
   return (
-    <span
+    <TagElement
       className={clsx(
         "inline-block bg-secondary text-body px-4 py-2 rounded-xl font-bold",
+        "underline hover:bg-primary hover:text-white transition-colors",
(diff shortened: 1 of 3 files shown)


patrick91:
thank you! I think we can remove the underline on the tags though, I don't think it looks good 😊

egeakman:
> thank you! I think we can remove the underline on the tags though, I don't think it looks good 😊

@patrick91 It's for accessibility's sake, to emphasize that they are links. But I'm OK with removing it.

patrick91:
@egeakman let's remove it for now, they look similarly to a button, so hopefully that's enough. But we'll also do a review for accessibility in future

src/components/tag/tag.tsx
```diff
@@ -1,30 +1,33 @@
...
+}
+
+export const Tag = ({ children, className, href }: TagProps) => {
+  const TagElement = href ? "a" : "span";
   return (
-    <span
+    <TagElement
       className={clsx(
         "inline-block bg-secondary text-body px-4 py-2 rounded-xl font-bold",
+        "underline hover:bg-primary hover:text-white transition-colors",
```
egeakman, line 15:
```suggestion
        "hover:bg-primary hover:text-white transition-colors",
```

PR #716, opened 2024-06-18
Add PyLadies Social Page

clytaemnestra:


cybit:
Can you please move the links around so that this order is followed
```
- Conference events by day, i.e. events that happen once, during the
conference, such as workshop, summit, etc.
- Social events by day, i.e. events that happen once outside of the main
conference.
- Multiple events / events that span several days, such as the PyLdadies
programme, hackathon, etc.
This is just a suggestion and there could be many more sensible way to
order this.
```

laisbsc (approved):
Awesome! TY <3

PR #719, opened 2024-06-19
Add missing events to /programme

Sangarshanan:
- Pyladies lunch 
- Community Organiser's Lunch
- Boat Cruise
- PyBurger


laisbsc (commented):
Looks good, just two more events

PR #742, opened 2024-06-25
Add job board from sponsor submissions

hypha:
FYI, I would have preferred to name the file "job board", instead of "jobs board". But that would have resulted in the sponsor info page link 404. 

Currently in draft, waiting for more submissions. Once ready, @aramboi and @laisbsc can review.


aramboi:
That page is already 404 so we could just change it to link to `/job-board` to begin with. ☺️

aramboi:
Awesome 🤩 LGTM :shipit:

laisbsc:
Looks awesome! @aramboi sent a reminder to all sponsors for the Job Board and Virtual Swag page and no one really replied. So might be better to just merge this and add more as we get it. If we get it.

Also, something it always intrigued me was why the Job Board and Virtual Swag are not under the Sponsor menu on the website 🤔 I can never find it under Info & Support.

hypha:
`Also something it always intrigued me was why the Job Board and Virtual Swag are not under the Sponsor menu on the website 🤔 I can never find it under Info & Support`.
Because attendees in the past said they can never find it under sponsorship 😅

laisbsc:
> `Also something it always intrigued me was why the Job Board and Virtual Swag are not under the Sponsor menu on the website 🤔 I can never find it under Info & Support`. Because attendees in the past said they can never find it under sponsorship 😅

I'm laughing out loud! 🤣  #howtomakeusershappy?

aramboi:
Maybe we should put the links in both submenus. I mean, I don't think it's unheard of, is it?
But for now, I'll merge like this and maybe add later. 😅

Issue #667, opened 2024-06-12
Talk rooms are not dispalyed on schedule in mobile view. 


hypha:
![image](https://github.com/EuroPython/website/assets/6319134/7f70b697-8ab9-4e32-bd37-330f8c7216a8)
looks like this on mobile. I think we need the rooms :blush:

egeakman:
cc @patrick91

hypha:
Thank you so much @patrick91! ❤️

PR #752, opened 2024-06-28
feeble attempt to display room names in mobile view

hypha:
An attempt to show room names in mobile view and resolve https://github.com/EuroPython/website/issues/667. 
Currently, it has a separate row instead of being inside the each session's grid. But it might still be readable. 
For registration sessions, the name is not displayed in mobile view as it is already in session title.


patrick91:
I've made a different implementation in #754, it's similar to last year 😊

PR #756, opened 2024-06-29
Fix LinkedIn profile link for speaker

goetzb:
I was just going through the schedule to decide which talks to attend and noticed a broken link on https://ep2024.europython.eu/speaker/michal-rokita to the LinkedIn profile of the speaker.

Not sure this way is how you want such suggestions and fixes to be reported, but I thought I'd give it a go as quick and easy option  without creating an issue or writing an email to the organiser.

### DIFF
--- a/src/content/speakers/GV9BKJ.mdx
+++ b/src/content/speakers/GV9BKJ.mdx
@@ -4,7 +4,7 @@ avatar: https://program.europython.eu/media/avatars/prof_lCHNRNw.jpg
 code: GV9BKJ
 gitx: https://github.com/mRokita
 homepage: null
-linkedin_url: "https://linkedin.com/in/linkedin.com/in/micha\u0142-rokita"
+linkedin_url: "https://linkedin.com/in/micha\u0142-rokita"
 mastodon_url: null
 name: "Micha\u0142 Rokita"
 slug: michal-rokita


egeakman:
Hi @goetzb!

Speaker, session, and schedule pages are automatically built and updated; so changing them here won't resolve the issue, it needs to be fixed in the source.

The data that is used to build these pages are created by [this project](https://github.com/EuroPython/programapi).

I'll look into why this happened. Thank you again for reporting the bug!

goetzb:
> Speaker, session, and schedule pages are automatically built and updated; so changing them here won't resolve the issue, it needs to be fixed in the source.
> 
> The data that is used to build these pages are created by [this project](https://github.com/EuroPython/programapi).

Thanks @egeakman, I had a suspicion that there's more to it than some Astro collection, but as I spotted it here, I thought it's worth a try - and you now know about it :)

egeakman:
> Thanks @egeakman, I had a suspicion that there's more to it than some Astro collection, but as I spotted it here, I thought it's worth a try - and you now know about it :)

Yes! Thanks for spotting it!

I opened a PR to fix the double ``linkedin.com`` in the URL: https://github.com/EuroPython/programapi/pull/52

egeakman:
It should be fixed now! Thanks for reporting :slightly_smiling_face:

PR #769, opened 2024-07-02
sprints: fix link to sprints page source

xrmx:
Currently getting a 404


cybit:
Thank you for the fix.

PR #770, opened 2024-07-02
Add redirect for discord

patrick91:
Looks like this is not officially supported by Astro, but might work


egeakman:
Thanks Patrick, you are the best!

PR #771, opened 2024-07-03
Changed ticket information for social event to sale open.

cybit:
Ticket sales are real now.


laisbsc (approved):
Nice and simple. Approved! ✅️

Issue #682, opened 2024-06-13
[A11y] Accessibility low-hanging fruits


Cheukting:
I use this tool (https://accessibilityinsights.io/) to run quick checks for simple issues, here is a list of them (not explicit):

[Main page]
- [ ] Color contracts are not enough for most of the buttons
- [ ] Quicklinks at the bottom - Ensures `<li>` elements are used semantically
- [ ] Logo on the top left and "Power by Vercel" btn at the bottom right missing link description

[Schedule]
- [ ] Date selection - Ensures select element has an accessible name
- [ ] The colour contrast at the "Advance" label is not enough

[List of speakers]
- [ ] Ensures that lists are structured correctly

If you would like to work on any of those, I recommend installing the same tool (https://accessibilityinsights.io/), it is a browser extension, that will guide you on how to fix them.

In general, there are not many issues that the tool found on our website. Of course, this is just a very simple tool and there maybe more we can do but these are low-hanging fruits.

nikoshell:
I think that we can close this issue.
We improve a11y in much more areas and we testing it to keep standards.
If there are any issues which are still not covered please create new ticket.

PR #773, opened 2024-07-03
A11y: fix red button contrast

hugovk:
For https://github.com/EuroPython/website/issues/682.

The WCAG AA min contrast ratio is 4.5.

# Before

contrast ratio: 3.99:1 https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=FF0000

image

# After

contrast ratio: 4.53:1 https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=EE0000

image


egeakman (approved):
Looks good to me. Thank you Hugo!

PR #774, opened 2024-07-03
A11y: fix card button contrast

hugovk:
For https://github.com/EuroPython/website/issues/682.

# Before

The contrast ratio of white on pink is too low: 2.22: 1

image

# After

Instead of white on pink, use white on read, like the other buttons on the front page. With https://github.com/EuroPython/website/pull/773, the ratio will be 4.53:1, above the WCAG AA minimum of 4.5.

image

### DIFF
--- a/src/components/card/card.astro
+++ b/src/components/card/card.astro
@@ -25,7 +25,7 @@ const { title, subtitle, url, image } = Astro.props;
   </div>
   <h3 class="text-3xl font-bold mt-6 mb-4 text-primary">{title}</h3>
   <p
-    class="bg-secondary inline-block text-text-inverted font-extrabold px-4 py-2 text-lg mb-4 rounded-[30px]"
+    class="bg-button inline-block text-text-inverted font-extrabold px-4 py-2 text-lg mb-4 rounded-[30px]"
   >
     {subtitle}
   </p>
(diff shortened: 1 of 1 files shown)


patrick91:
@hugovk thanks, but I'd prefer to keep the pink if possible, the red is too in-your-face :D 

maybe we can find another shade of pink we can use?

hugovk:
Would you like to have a play with the colour picker at https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=FF8B9D and see if there's one you like that meets WCAG AA for normal text (4.5:1)?

src/components/card/card.astro
```diff
@@ -25,7 +25,7 @@ const { title, subtitle, url, image } = Astro.props;
   </div>
   <h3 class="text-3xl font-bold mt-6 mb-4 text-primary">{title}</h3>
   <p
-    class="bg-secondary inline-block text-text-inverted font-extrabold px-4 py-2 text-lg mb-4 rounded-[30px]"
+    class="bg-button inline-block text-text-inverted font-extrabold px-4 py-2 text-lg mb-4 rounded-[30px]"
```
patrick91, line 28:
```suggestion
    class="bg-[#e22558] inline-block text-text-inverted font-extrabold px-4 py-2 text-lg mb-4 rounded-[30px]"
```

I think this doesn't look bad 😊

egeakman (approved):
Let's merge this!

PR #806, opened 2024-07-09
Update aitournament.mdx

nvaytet:
Because of various reasons (logistical among others), I made the repositories public a little earlier, and the challenge is now open.


egeakman (approved):
Thanks!

PR #833, opened 2024-07-11
Add CPython to sprints

ambv:


rodrigogiraoserrao (approved):
Thanks for updating room occupancy. People always overlook that 🙃

PR #907, opened 2024-10-01
Fix the footer link to point to the 2024 photos instead of 2023

cybit:
The current footer still points to the 2023 photos instead of the 2024 photos we have now.

### DIFF
--- a/src/data/links.json
+++ b/src/data/links.json
@@ -182,8 +182,8 @@
   ],
   "footer": [
     {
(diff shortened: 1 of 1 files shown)


cybit:
Just want to have a second pair of eyes on this.

laisbsc (approved):
Look good. Thank ya :)

egeakman (approved):

PR #916, opened 2024-10-15
Add YouTube embeds to session pages

egeakman:


cybit (approved):
Looks good to me. 
At the first moment I was looking for a link to the full playlist. But we do have this in the header anyway.
And I like the embedded way of presenting the video instead of clicking through.

hypha (approved):
All looks good to me. Thank you so much, Ege! All dogs are belong to you! <3

PR #929, opened 2024-12-17
Add Bluesky to contact page

hugovk:
And move X to the end (personally I'd recommend removing it altogether).


egeakman:
Hey Hugo, it's nice to see you here! Thank you!

Don't mind the Vercel build failure, we're in the process of migrating out of Vercel because their sponsorship has expired. We are trying to centralize our infrastructure and soon we will have our own previews 🙂

egeakman (approved):

PR #940, opened 2025-01-10
Add CfP CTA, and Redirection

egeakman:
- Adds a CTA for Call for Proposals
- Add a redirection for `/programme/cfp` -> `#` (change to actual link)

Contributes to EuroPython/ep2025#46

Also cc @t-redactyl


src/content/deadlines/02_cfp.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Call for Proposals
+subtitle: Opens Until 27 January
```
egeakman, line 3: This might sound like it is opening on 27 January, instead of ending.

hypha: Hmmm doesn't sound like that to me. But to avoid confusion, how about Opens Now till 27 January, or just put the whole date range "Opens 10-27 January"?

hypha: Sure that works too! I'll get that in just now.

PR #943, opened 2025-01-10
Add initial version for mentorship programme page

cmaureir:
This is a first version for the page, that does not include forms nor dates


src/data/links.json
```diff
@@ -9,6 +9,10 @@
         }
       ]
     },
+    {
+      "name": "Mentorship",
+      "path": "/mentorship"
+    },
```
egeakman, line 15: Do we not want this under Programme? With the URL `/programme/mentorship`?

cmaureir: good question, even though the mentorship program is a different team, it's related to the programme for web consumers. Let me update it

cmaureir:
Naa gave a :+1: on Discord, so if there is no further comment, I believe we can merge this.

egeakman (approved):
Lgtm

PR #961, opened 2025-01-15
Add deployment automation

egeakman:
- This does not include clearing the temp directory. We should set a cron job for that.
- Also the previews directory might need some cleaning too after a while, but I guess that's something to discuss elsewhere.
- The files are copied to a temp directory from local machine (or runner), then copied to their final destination (within the static server). This is as atomic as I could get it (and most likely should be enough).


deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+    branch_name:
+      "{{ lookup('pipe', 'git rev-parse --abbrev-ref HEAD') | trim }}"
+    sanitized_branch_name:
+      "{{ branch_name | regex_replace('[^a-zA-Z0-9_-]', '_') }}"
+    remote_base_path: "/home/{{ ansible_user }}/content"
+
+  tasks:
+    - name: Generate UUID for temp directory
+      set_fact:
+        unique_id: "{{ lookup('pipe', 'uuidgen') }}"
```
cybit: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/to_uuid_filter.html seems to work without leaving ansible. Use the time to randomize output: `{{ ansible_date_time.iso8601_micro | to_uuid }}`
Or `{{ 99999999 | random | to_uuid }}`
(See https://stackoverflow.com/questions/30516011/ansible-random-uuid-generation)

deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+          {{
+            (
+              remote_base_path + '/europython_websites/' + branch_name
+              if branch_name in branch_list else
+              remote_base_path + '/previews/' + sanitized_branch_name
+            ) | trim
+          }}
+        temp_dir:
+          "{{ remote_base_path }}/temp/temp_{{ sanitized_branch_name }}_{{
+          unique_id }}"
```
cybit:
```suggestion
          "{{ remote_base_path }}/temp/{{ unique_id }}"
```
This is the temp directory. Which goes against all unix conventions so it can not be from the system. And the only manual use would be for the rare edge case where copy was successive but the move would not work. In which case `stat` information (creation timestamp) should be enough.

And why do we use this? https://docs.ansible.com/ansible/latest/collections/ansible/posix/synchronize_module.html#parameter-delay_updates is the default ?

deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+          - "Branch Name: '{{ branch_name }}'"
+          - "Sanitized Branch Name: '{{ sanitized_branch_name }}'"
+          - "Remote Base Path: '{{ remote_base_path }}'"
+          - "Deploy Path: '{{ deploy_path }}'"
+          - "Temp Dir: '{{ temp_dir }}'"
+
+    - name: Ensure the temporary destination directory exists
+      file:
+        path: "{{ temp_dir }}"
+        state: directory
```
cybit:
```suggestion
        state: directory
        recurse: true
        mode: 750
```

deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+      debug:
+        msg:
+          - "Branch Name: '{{ branch_name }}'"
+          - "Sanitized Branch Name: '{{ sanitized_branch_name }}'"
+          - "Remote Base Path: '{{ remote_base_path }}'"
+          - "Deploy Path: '{{ deploy_path }}'"
+          - "Temp Dir: '{{ temp_dir }}'"
+
+    - name: Ensure the temporary destination directory exists
+      file:
```
cybit:
```suggestion
      ansible.builtin.file:
```

deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+      "{{ branch_name | regex_replace('[^a-zA-Z0-9_-]', '_') }}"
+    remote_base_path: "/home/{{ ansible_user }}/content"
+
+  tasks:
+    - name: Generate UUID for temp directory
+      set_fact:
+        unique_id: "{{ lookup('pipe', 'uuidgen') }}"
+
+    - name: Define and trim deploy path
+      set_fact:
```
cybit:
```suggestion
      ansible.builtin.set_fact:
```

deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+              if branch_name in branch_list else
+              remote_base_path + '/previews/' + sanitized_branch_name
+            ) | trim
+          }}
+        temp_dir:
+          "{{ remote_base_path }}/temp/temp_{{ sanitized_branch_name }}_{{
+          unique_id }}"
+
+    - name: Show deployment configuration
+      debug:
```
cybit:
```suggestion
      ansible.builtin.debug:
```

deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+      - ep2025
+    branch_name:
+      "{{ lookup('pipe', 'git rev-parse --abbrev-ref HEAD') | trim }}"
+    sanitized_branch_name:
+      "{{ branch_name | regex_replace('[^a-zA-Z0-9_-]', '_') }}"
+    remote_base_path: "/home/{{ ansible_user }}/content"
+
+  tasks:
+    - name: Generate UUID for temp directory
+      set_fact:
```
cybit:
```suggestion
      ansible.builtin.set_fact:
```

deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+          - "Deploy Path: '{{ deploy_path }}'"
+          - "Temp Dir: '{{ temp_dir }}'"
+
+    - name: Ensure the temporary destination directory exists
+      file:
+        path: "{{ temp_dir }}"
+        state: directory
+
+    - name: Synchronize local files to remote temporary directory
+      synchronize:
```
cybit:
```suggestion
      ansible.posix.synchronize:
```
Are you sure ansible.posix is part of your environment? I guess not if you are installing ansible via pip. 

```
To install it, use: ansible-galaxy collection install ansible.posix.
To use it in a playbook, specify: ansible.posix.synchronize.
```
(https://docs.ansible.com/ansible/latest/collections/ansible/posix/synchronize_module.html)

deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+        src: dist/
+        dest: "{{ temp_dir }}/"
```
cybit:
```suggestion
        dest: "{{ deploy_path }}/"
```
https://docs.ansible.com/ansible/latest/collections/ansible/posix/synchronize_module.html#parameter-delay_updates (true by default) should take care of any sync problems.

deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+        mode: push
```
cybit:
```suggestion
        mode: push
        checksum: true 
```
Optional. will compare the checksums and not overwrite files that have not been changed. 
(https://docs.ansible.com/ansible/latest/collections/ansible/posix/synchronize_module.html#parameter-checksum)

deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+        rsync_opts:
+          - "--delete"
```
cybit:
```suggestion
        delete: true
```
Optional. - This is not exactly the same as your --delete. It does add --delete-after. So it will transfer first and then delete.
With the temp dir --delete would not have made any sense as the dest-dir would have been new every time.

deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+
+    - name: Ensure temporary directory is populated
+      shell: test "$(ls -A {{ temp_dir }})"
+      register: temp_dir_check
+      failed_when: temp_dir_check.rc != 0
+
+    - name: Copy files from temporary directory to deployment directory
+      shell: rsync -a --delete {{ temp_dir }}/ {{ deploy_path }}/
+
```
cybit:
```suggestion
```
[delay-updates](https://docs.ansible.com/ansible/latest/collections/ansible/posix/synchronize_module.html#parameter-delay_updates) should take care of this.
If you really want to make sure that the sync works you do not want to do a local rsync eiter. 

The way to do this on linux would either be: `rm {{ deploy_path }} && mv {{ temp_dir }} {{ deploy_path }}`
OR 
not having the files in the location at all and keep the temp_dirs instead and then `ln -sf {{ temp_dir }} {{ deploy_path }}` .  But that can have implications on the webserver configuration.

deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+    - name: Verify deployment success
+      shell: |
+        if [ ! -d "{{ deploy_path }}" ] || [ -z "$(ls -A {{ deploy_path }})" ]; then
+          exit 1
+        fi
```
cybit:
```suggestion
      ansible.builtin.find:
        paths: "{{ deploy_path }}"
        recurse: true
        age: -1d
```
Do not use shell if you don't have to.
recurse together with age is optional. The idea here is that we check if there are any new files in the directories.

deploy.yml
```diff
@@ -0,0 +1,72 @@
...
+      register: target_dir_check
+      failed_when: target_dir_check.rc != 0
```
cybit:
```suggestion
      failed_when: target_dir_check.matched == 0
```
Find will only find [files](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/find_module.html#parameter-file_type) in this config. So even updated directories would not be found.

.github/workflows/build-deploy.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: pnpm install
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+          cache: "pip"
+
+      - name: Install Ansible
+        run: pip install ansible
```
cybit: Code below needs  `ansible-galaxy collection install ansible.posix.` to install the `synchronize` module.

cybit (changes requested):
A lot is linting. 
Some is don't reinvent. 
Some is missing.

deploy.yml
```diff
@@ -0,0 +1,64 @@
...
+    branch_name:
+      "{{ lookup('pipe', 'git rev-parse --abbrev-ref HEAD') | trim }}"
+    sanitized_branch_name:
+      "{{ branch_name | regex_replace('[^a-zA-Z0-9_-]', '_') }}"
+    remote_base_path: "/home/{{ ansible_user }}/content"
+
+  tasks:
+    - name: Generate UUID for temp directory
+      ansible.builtin.set_fact:
+        unique_id: "{{ lookup('pipe', 'uuidgen') }}"
```
cybit:
```suggestion
        unique_id: "{{ ansible_date_time.iso8601_micro | to_uuid }}"
```

.github/workflows/build-deploy.yml
egeakman: We install ansible-core from apt then

src/components/footer.astro
```diff
@@ -89,5 +92,7 @@ import { EPSLogo } from "./logo/eps-logo";
         </p>
       </div>
     </article>
+
+    <p class="mb-4" style="color: rgba(255, 255, 255, 0.4)">version: {gitVersion} @ {buildTimestamp}</p>
```
cybit:
```suggestion
          <p class="mb-4" style="color: rgba(255, 255, 255, 0.4)">version: {gitVersion} @ {buildTimestamp}</p>
      </div>
    </article>

```

cybit: Was better but I want to be able to find it. :)

cybit (approved):
One thing about moving the version information directly beneath the links in the footer. But that is not neccessary.

src/components/footer.astro
artcz: Yes, I think I messed up the previous merge and only did the mb-4 without moving the tag :D

PR #966, opened 2025-01-20
add previews

artcz:


.github/workflows/preview.yml
```diff
@@ -0,0 +1,102 @@
...
+
+on:
+  pull_request:
+
+jobs:
+  preview:
+    name: Run preview
+    runs-on: ubuntu-latest
+    env:
+      PREVIEW_HOSTNAME: ep-preview.click
```
cybit, line 11: Could we have this as an environment variable please? Makes it reusable for others.
```suggestion
      PREVIEW_HOSTNAME: ${{ vars.PREVIEW_HOSTNAME }}
```

README.md
```diff
@@ -70,3 +70,5 @@ import image from "./image.jpg";
 
 <Image src={image} width={500} />;
 ```
+
+.
```
cybit:
```suggestion

```
Why is there a dot at the end? I do understand that maybe a \n\n is wanted.

src/components/footer.astro
```diff
@@ -93,6 +93,6 @@ const gitVersion = import.meta.env.GIT_VERSION;
       </div>
     </article>
 
-    <p>version: {gitVersion} @ {buildTimestamp}</p>
+    <p style="color: rgba(255, 255, 255, 0.4)">version: {gitVersion} @ {buildTimestamp}</p>
```
cybit:
```suggestion
    <p class="mb-4" style="color: rgba(255, 255, 255, 0.4)">version: {gitVersion} @ {buildTimestamp}</p>
```
And please move this up behind 
```
            github.com/europython <span> ↗</span>
          </a>
        </p>
```
and in front of
```
      </div>
    </article>
```

cybit (approved):
Mostly cosmetics

.github/workflows/preview.yml
artcz: Good point! - but I'm going to keep it as-is for now, because there are other things in that GHA that are enviornment-specific, so might be better to just have a separate commit/PR dealing with all those changes at once.  – for example there's a separate deployment hostname (static.ep).
Separately it's probably a good idea to structure it as separate build/preview/deploy steps, so could be done as part of that change. Currently deploy/preview share a lot of logic that can be put in a separate "build" step.

PR #979, opened 2025-01-26
Update CfP deadline after extension

cmaureir:


cmaureir:
Official announcement goes out tomorrow 8am CET. We need to hopefully deploy at the same time

Sangarshanan (approved):

PR #998, opened 2025-02-06
fix typo

artcz:


hypha (approved):
Woo! Verbs are important! Thanks!

PR #1000, opened 2025-02-06
Add community voting CTA

hypha:


hypha:
hmmm the preview is not working. I wonder what I've done wrong here.

hypha:
the issue was using uppercase in branch name. New branch with lowercase worked here: https://github.com/EuroPython/website/pull/1004 closing this. 

Unfortunately, this will not be quite useless as a demo for how uppercase branch name doesn't have correct preview, because https://ep2025-voting-cta.ep-preview.click/ is no longer 404 after the 1004 PR.

PR #1007, opened 2025-02-06
Add link to community voting to the header.

cybit:
Closes https://github.com/EuroPython/ep2025/issues/68

### DIFF
--- a/src/components/header/header-actions.astro
+++ b/src/components/header/header-actions.astro
@@ -21,11 +21,9 @@ const IS_LIVE = false;
           </abbr>
           <span class="hidden md:inline">Code of Conduct</span>
         </ButtonLink>
-<!--
-        <ButtonLink href="/programme/cfp" className="hidden md:block">
-          <span>Send your proposal!</span>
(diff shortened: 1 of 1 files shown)


clytaemnestra:
I tried all links related to community voting and all of them LGTM!

clytaemnestra (approved):

hypha:
Everything looks good. I just wonder if it is a good idea to link directly to the voting page or better linking it to https://ep2025.europython.eu/voting. I think years ago @artcz was interested in gathering the analytics and not sure if direct voting page has that.

artcz:
yes, thanks noticing @hypha ! The ~~/voting analytics correctly, so it should be all ok – although hard to say now if the source/referer is preserved.~~ nvm, I re-read the PR – but that still holds – voting has it's own plausible :sweat_smile: 

That said the voting app doesn't contain the same information as the website, but should still be good enough to vote.

What I noticed while testing it is that ublock seems to by default block plausible so it's quite possible that our analytics are lower than real page views already :sweat_smile:

PR #1008, opened 2025-02-10
Add disabled prop to ButtonLink

### LINKED ISSUE #978: Add disabled prop to Button and ButtonLink
Add a `disabled` prop for Button and ButtonLink. Make it gray and unclickable.

*Inspired from @hypha at https://github.com/EuroPython/website/pull/976#discussion_r1928728683*

egeakman:
- Migrates ButtonLink to Astro, and adds a disabled prop to it.
- Removes redundant ButtonWithTitle component.

See https://add-disabled-prop.ep-preview.click/programme/mentorship/ for example

Closes #978


src/components/card/card.astro
```diff
@@ -28,7 +28,7 @@ const { title, subtitle, url, image } = Astro.props;
     <h3 class="text-3xl font-bold mt-2 mb-4 text-primary">{title}</h3>
```
hypha, line 28: I think the h3 is overwriting the tailwind font size. I am not sure why we need the h3 here. But if we keep it, we need to use `class="!text-3xl` to force overwrite.

src/components/button-link/button-link.astro
```diff
@@ -0,0 +1,32 @@
...
+const { url, className = "", secondary = false, disabled = false, isExternal} = Astro.props;
+const resolvedIsExternal = isExternal !== undefined ? isExternal : url?.startsWith("http");
+---
+
+<a
+  class={`font-bold text-lg px-4 py-4 bg-button rounded-[60px] inline-block leading-4 hover:bg-button-hover not-prose
+    ${secondary ? "bg-primary text-white hover:bg-primary-hover" : "text-text-inverted"}
+    ${secondary ? "text-text" : ""}
+    ${className || ""}
+    ${disabled ? "opacity-50 pointer-events-none" : ""}`}
```
hypha, line 22: I am a bit concerned that the contrast is too low. I think it might be better to use the secondary button styling for the disabled, e.g. changing the line 19 to `${secondary || disabled ? "bg-primary text-white hover:bg-primary-hover" : "text-text-inverted"}`? I think a greyed out button looks better too (and hopefully, the secondary colour will always be a fairly dark one in the future). WDYT?

hypha (commented):
Thanks @egeakman! I was a bit scared when the email notification says 14 files changed for the ButtonLink PR.  But now I see you've done a much bigger job! <3 

Questions not in diff: 
1. The only place we still use the buttons are in the voting.mdx page. I am not sure if it makes sense to keep the component just for that. 
2. could we delete the card-container.tsx file now? 

A small comment: I think I need to speed up the _test.mdx now so people don't get frustrated when href no longer works in ButtonLink :D

egeakman:
> 1. The only place we still use the buttons are in the voting.mdx page. I am not sure if it makes sense to keep the component just for that. 

It is a basically a different looking button-ish component which is not really my favorite. I think it was also used in the finaid page.

> 2. could we delete the card-container.tsx file now? 

I was tempted to convert some more components but thought maybe those would be more appropriate for separate PRs. We can remove it and do the card placement logic in the index, or keep it and convert it to Astro.

src/components/card/card.astro
hypha: Ah, this one is actually from the past. So perhaps we can address it in another PR when we look at the card component later?

hypha:
> It is a basically a different looking button-ish component which is not really my favorite. I think it was also used in the finaid page.

Agreed. I would like to dump it too. And since this is the PR related to buttons, maybe we can address it here? I don't see any use of this button component from finaid. Seeing that the styling in the voting app drifted a lot now from the current site, I think it is pointless to keep this button here. Instead we can just use bullet points here in the voting page?

> I was tempted to convert some more components but thought maybe those would be more appropriate for separate PRs. We can remove it and do the card placement logic in the index, or keep it and convert it to Astro.

You are right! Let's address the card in another PR! This can be resolved too: https://github.com/EuroPython/website/pull/1008#discussion_r1949623818

src/components/card/card.astro
egeakman: Yeah, let's address it in another PR. Thanks for pointing it out!

egeakman: Created an issue to track this #1011

hypha (approved):
Thank you for addressing all the comments, Ege! My brain wouldn’t have gone to such a nice and simple solution like using a screenshot for the vote options. <3

PR #1031, opened 2025-03-09
Implement a redirection system with a mapping

### LINKED ISSUE #1012: Set up redirect mapping
> Also, would it be good to set up a redirect in case we want to promote this on social media?

Yeah, we can set up a redirect. But to be honest I wasn't expecting to have this number of redirects, so it may be better to set up a mapping in the code instead of a page for each redirect.

For now we can just create that redirection page, and I'll look into how we can do it with a mapping later.

_Originally posted by @egeakman in https://github.com/EuroPython/website/issues/995#issuecomment-2646358439_

egeakman:
Redirections are now defined in `src/data/redirects.json`

Closes #1012


artcz (approved):
Not an astro expert, and this seems better than what we have currently so LGTM :)

I would like us to figure out at some point how to do it server-side, but it might not be that easy in our setup, and definitely very low priority. Let's focus on other things first.

Thanks for putting it together :)

hypha:
Thanks so much for working on this Ege! Being an egg to pretty much everything related to web, I don't understand the details of the code well enough, but do get the intention (which looks good to me) and the ep-preview looks good too. But I get "connection reset" when trying to preview locally (I use docker). Maybe I did something wrong, but I want to confirm the docker build still works if this gets merged. 

Also, might it be helpful to put a wee paragraph in the readme on how to update the redirect file? <3

nikoshell:
Please consider to use Astro build in redirects.
https://docs.astro.build/en/guides/routing/#redirects

egeakman:
> Please consider to use Astro build in redirects. https://docs.astro.build/en/guides/routing/#redirects

For some reason, I thought this didn't work in a static environment in the past. It looks promising, thank you!

PR #1039, opened 2025-03-11
Migrate tailwindcss to v4, and use it through vite plugin

### LINKED ISSUE #1038: Migrate tailwindcss to v4, and use vite it through vite plugin
Image

egeakman:
[The `hidden` class no longer seems to work along with other classes that affect display.](https://github.com/tailwindlabs/tailwindcss/issues/15884)

I had to use a workaround like this:

```astro
${className.includes("hidden") ? "hidden" : "inline-block"}
```

Also, there seems to be a new preferred way of configuration: inline configuration in CSS. I like it.

I had to tweak some components and pages to make this work, but I most likely missed something somewhere.

Closes #1038


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://migrate-tailwind.ep-preview.click|
|last update|2025-03-11T05:55:30.427Z|

egeakman:
@hypha This will likely create conflicts with the sponsor PR because the colors were defined in the tailwind config. How would you like to move forward with this 🤗

egeakman:
Let's hold this until the sponsor PR gets merged!

egeakman:
It is probably the safest option to hold this until after the conference, since it might affect existing and new pages.

PR #1067, opened 2025-03-26
Update card component look.

nikoshell:
🖼️ Preview available 🖼️ : https://ep-website--1067.org.readthedocs.build/

### DIFF
--- a/src/components/card/card.astro
+++ b/src/components/card/card.astro
@@ -13,7 +13,7 @@ export interface Props {
 const { title, subtitle, url, image } = Astro.props;
 ---
 
-<div class="block min-w-[250px] w-full md:w-[40%] lg:w-[30%] p-4 bg-[#D5D6E6] rounded-2xl text-center flex flex-col h-full">
+<div class="block min-w-[250px] w-full md:w-[40%] lg:w-[30%] p-4 bg-[#D5D6E6] rounded-2xl text-center flex flex-col">
   <a href={url} class="block">
     <div class="relative w-full mb-4">
       <Image
@@ -26,7 +26,7 @@ const { title, subtitle, url, image } = Astro.props;
       />
     </div>
   </a>
-  <div class="flex flex-col items-center flex-grow mb-2">
+  <div class="flex items-center flex-col flex-grow mb-2">
     <a href={url} class="block hover:underline">
       <h3 class="text-2xl text-primary">{title}</h3>
     </a>
--- a/src/components/card/card.astro
+++ b/src/components/card/card.astro
@@ -13,7 +13,7 @@ export interface Props {
 const { title, subtitle, url, image } = Astro.props;
 ---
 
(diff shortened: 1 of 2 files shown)


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-updates-section.ep-preview.click|
|last update|2025-03-27T07:04:45.569Z|

src/components/card/card.astro
```diff
@@ -13,28 +13,28 @@ export interface Props {
...
-    className="mb-4"
-    isExternal={false}
-  >
+  <div class="flex flex-col items-start">
+    <a href={url} class="block hover:underline">
+      <h3 class="text-2xl text-primary">{title}</h3>
+    </a>
+      <slot />
+  </div>
+<a class="mt-2 text-xs px-2 py-1 bg-[#141F36] text-white rounded-[10px] inline-block leading-4 hover:bg-button-hover not-prose rounded-[5px]" href="{url}">
```
clytaemnestra: @nikoshell thanks for the MR! I'm aware this is based on the Figma mock and looks fine on desktop, but I checked the mobile version and have two suggestions.


When cards are stacked horizontally on mobile, the content is on the left side. 
image

I'd center the content inside the container, it looks more balanced to me: 
image


When the text inside the card is short, the button floats higher. 
image


I'd add a fixed top margin to the button. 
image


I tried these changes locally and here’s the updated card with my suggestions, so you don't have to do it from scratch: 
```
<div class="block min-w-[250px] w-full md:w-[40%] lg:w-[30%] p-4 bg-[#D5D6E6] rounded-2xl text-center flex flex-col">
  <a href={url} class="block">
    <div class="relative w-full mb-4">
      <Image
        src={image}
        width={600}
        height={375}
        alt=""
        class="rounded-xl aspect-[3/2] w-full"
        loading="lazy"
      />
    </div>
  </a>
  <div class="flex flex-col items-center">
    <a href={url} class="block hover:underline">
      <h3 class="text-2xl text-primary">{title}</h3>
    </a>
      <slot />
  </div>
<a class="mt-2 text-xs px-2 py-1 bg-[#141F36] text-white rounded-[10px] inline-block w-fit mx-auto leading-4 hover:bg-button-hover not-prose rounded-[5px] mt-auto" href="{url}">
    {subtitle}
 <span class="inline-block ml-1 font-system text-lg leading-4" >
↗
</span></a>
</div>
```

What do you think about these suggestions?

nikoshell: Fixed.

clytaemnestra: Looks a bit broken in the mobile version now - I think there is an additional property or something that broke the layout - it worked today correctly with the code from above 

image

nikoshell: Fixed.

clytaemnestra (approved):
Thanks, LGTM!

PR #1083, opened 2025-03-30
Add manual trigger to RTD preview, run workflow on forks.

nikoshell:
🖼️ Preview available 🖼️ : https://ep-website--1083.org.readthedocs.build/

### DIFF
--- a/.github/workflows/rtd-preview.yml
+++ b/.github/workflows/rtd-preview.yml
@@ -10,6 +10,7 @@ permissions:
 
 jobs:
   documentation-links:
+    if: github.event.pull_request.head.repo.fork == true
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
--- a/.github/workflows/rtd-preview.yml
+++ b/.github/workflows/rtd-preview.yml
@@ -3,6 +3,7 @@ on:
   pull_request_target:
     types:
       - opened
+  workflow_dispatch:
 
 permissions:
   pull-requests: write


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep20252-preview.ep-preview.click|
|last update|2025-04-01T11:32:55.845Z|

.github/workflows/rtd-preview.yml
```diff
@@ -3,12 +3,14 @@ on:
...
     types:
       - opened
+  workflow_dispatch:
 
 permissions:
   pull-requests: write
 
 jobs:
   documentation-links:
+    if: github.event.pull_request.head.repo.fork == true
```
egeakman, line 13: We were talking about entirely moving to RTD for previews. I think we can just remove `preview.yml`.

@artcz thoughts?

nikoshell: RTD is failing right now with undefined reason and it's blocking few PR.
this change keep current preview which is working fine for pull requests and use RTP for forks.

artcz: Yes, some of the builds are failing – the speaker one is because of the pipx missing (which maybe we should migrate to uvx :D), but the other one doesn't have an obvious reason in the output.

I think it might be a good idea to keep RTD preview just for forks, because that way we also don't overload the RTD servers with unnecessary previews, plus we're testing closer to the same setup as production if we keep the same preview flow. 

Looking at our server, so far we generated ~350 previews totalling 6.5G in the last ~two months.

artcz (approved):

PR #1084, opened 2025-03-30
Add sessions preview page

cmaureir:
Sessions are already being confirmed, so we can have a preliminary preview page with a list of them

🖼️ Preview available 🖼️ : https://ep-website--1084.org.readthedocs.build/

### DIFF
--- a/src/content/pages/programme/sessions.mdx
+++ b/src/content/pages/programme/sessions.mdx
@@ -93,4 +93,3 @@ Go to session type:
 ## Poster
 
 * "Psychological Model for Mapping and Prediction of Stress Among Students" by Oladapo Kayode Abiodun, Oluwakemi Jacobs
-
--- a/src/content/pages/programme/sessions.mdx
+++ b/src/content/pages/programme/sessions.mdx
@@ -1,0 +1,96 @@
+---
+title: Sessions Preview EuroPython 2025
+subtitle: This is a preliminary list of the confirmed sessions for EuroPython 2025. The list will get modified until the final schedule is published.
+---
+
+# Sessions preview
+
+The following sessions are a preliminary list of the confirmed proposals that
+will be part of the conference. The list will keep growing in the following
+days and some sessions might be replaced in case of them being withdrawn by the
+authors.
+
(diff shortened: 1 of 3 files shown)


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://sessions-preview.ep-preview.click|
|last update|2025-04-02T19:51:32.843Z|

cmaureir:
We cannot use URLs for pretalx sessions nor speakers because we don't have a published schedule.
Would it be OK to do it like this or should we maybe add the abstract information as well? (maybe on a `<details>` ?)

hypha:
Thanks so much @cmaureir! I think this is a really good temporary solution to push out the list of sessions by the communicated deadline! 
- Maybe it will be slightly more readable if you make each title bold, like **“Mastering decorators: the cherry on top of your functions”** by Rodrigo Girão Serrão
- imho, having the abstract will definitely be nicer. But if you need to update it frequently, it could become more of a chore? If you decide to do it, you could also consider using the accordion component. See here: https://ep2025.europython.eu/test/#accordion

src/content/pages/programme/sessions.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+# Sessions preview
+
+The following sessions are a preliminary list of the confirmed proposals that
+will be part of the conference. The list will keep growing in the following
+days and some sessions might be replaced in case of them being withdrawn by the
+authors.
+
+Go to session type:
+* [Keynotes](#keynote)
```
thepetk, line 14: Just to confirm that this is intended: "Keynotes" (plural) <> "Keynote" (singular)

thepetk (commented):
Lgtm in general, I agree on @hypha point about making the title bold, it could make the text more readable.

clytaemnestra (approved):
LGTM!

src/content/pages/programme/sessions.mdx
cmaureir: Yeah, it's just an anchor to the section that was called Keynote. I used Singular in all the types :|

PR #1105, opened 2025-04-03
Update md headlines.

nikoshell:
image


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-headlines.ep-preview.click|
|last update|2025-04-03T21:35:58.799Z|

src/styles/global.css
```diff
@@ -52,6 +52,24 @@ body {
...
+  transition:
+    color 0.2s ease-in-out,
+    background-color 0.2s ease-in-out,
+    text-decoration-thickness 0.2s;
+  color: #151f38;
+}
+
+.prose :is(h1, h2, h3, h4, h5, h6) a:focus {
+  outline: 4px solid #0073e6;
+  outline-offset: 5px;
```
clytaemnestra: looks funny, but I like it! can we update the `--color-primary-active` as well? `#3c3ffa` is from the 2023 design system. 

image

nikoshell: outline color was chooses to get high contrast & accessibility it's also standardized and users are familiar with it

If i understand correctly changing --color-primary-active will affect  whole website and it's not part of this PR

clytaemnestra: the outline color `#3c3ffa` doesn't look nice in this design system, though - is it possible to change it to `#151f38`? it can be done in a separate PR later.

src/styles/global.css
```diff
@@ -52,6 +52,22 @@ body {
...
   text-decoration: none !important;
+  background-image: linear-gradient(45deg, #151f38, #7a7ebd);
+  color: transparent;
+  background-clip: text;
+  -webkit-background-clip: text;
+}
+
+.prose :is(h1, h2, h3, h4, h5, h6) a:hover {
+  text-decoration: underline !important;
+  transition: color 0.2s ease-in-out, background-color 0.2s ease-in-out, text-decoration-thickness 0.2s;
```
cybit: Please keep in mind that there are people out there that have problems with contrast. And as nice as the fading might seem - it does make for a very bad contrast.

If you really need this then we should at least add a very visible and easy to use option to change to a different styleset that does use high contrast only.

Screenshot 2025-04-03 at 16 20 49

cybit (changes requested):

src/styles/global.css
clytaemnestra: @nikoshell can you please try to improve the accessibility? if not, I'd say let's keep the old style. thank you!

hugovk: It is distracting, I think it's better to keep it "boring" and functional.

egeakman: I agree with Cyril and Hugo on this one.

nikoshell: Colors are just suggestion as we don't have design.
I can use AAA standards, if you force it as priority.

Let me suggest better colors.

About other warnings from Lighthouse, this PR doesn't fix it, we also use third party to convert markdown to html so first we need to be sure that this is our issue or issue which is our priority.

cybit (approved):
Thank you. My concern about the fade has been addressed.

While we are at it we should look into the contrast at mouse-over as well. But I guess that has been there before this PR.

egeakman (approved):
Thank you!

PR #1111, opened 2025-04-04
Improve search.

### LINKED ISSUE #1052: Search on Mobile
Now that Search #1049 is merged – we should look into making it available on mobile screens as well :)

nikoshell:
Close #1052.


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-search.ep-preview.click|
|last update|2025-04-15T15:49:41.154Z|

clytaemnestra (commented):
this doesn't look quite well on the desktop

image

egeakman:
Can we keep the old search bar for bigger screens?

nikoshell:
There is a place for suggestions feature.

nikoshell:
Search should be inside of modal but we don't have that component implemented yet.

nikoshell:
> Can we keep the old search bar for bigger screens?

I think no we don't want have two different components for the same functionality, current one is unified for mobile and desktop, and before we had responsiveness issues. When we will have modal implemented desktop search can be inside modal.

clytaemnestra:
@nikoshell so we have two options:
1. either have this, the same, component for both mobile and desktop, which doesn't look nice, but works
2. or have no search on mobile at all 

Do I understand it correctly?

nikoshell:
I will improve functionality to looks like this one https://starlight.astro.build/

nikoshell:
There are header changes which fix this issue.
Close #1120.

src/components/Search.astro
```diff
@@ -0,0 +1,154 @@
...
+
+  <div class="search-header flex items-center justify-between z-40 h-0 md:h-[85px] m-auto ">
+    <h2 class="text-black text-4xl ml-10 top-[46px] md:top-[85px]">Search<h2/>
+    <Button
+      id="search-close"
+      clear
+      icon="close"
+      iconSize="fa-xl"
+      class="w-[3em] h-[3em] top-[46px] md:top-[85px]"
+    />
```
clytaemnestra: Lighthouse shows me that the close button is missing an accessible name - can it be added? 

image

clytaemnestra (commented):
Looks and works great, thanks!

src/components/Search.astro
clytaemnestra, line 1: Can we change the text color to black? This brown doesn't fit with the current colors 

image

egeakman:
The upper part (Search text and close button) looks a bit weird when scrolled down. Maybe we can add some sort of semi-transparent background to the top area (something similar to the ctrl+k text)?

image

egeakman:
Can we make the search pop-up lose focus (close) when clicked outside of it? Also, the search menu closes when I hit search on the phone keyboard.

src/components/Search.astro
nikoshell: Please re-verify. I was checking Lighthouse and can't reproduce after changes.

clytaemnestra: LGTM

clytaemnestra (approved):

egeakman (approved):
Looks and works good!! Thanks!

PR #1135, opened 2025-04-12
Short links.

### LINKED ISSUE #1002: Add an interim page to the website for menue items.
We should add a page to the website that mirrors the menue. 
Now: If I go to ep2025.europython.eu/programme I get a 404 .
Expected outcome: I get either a defined page explaining stuff or I get a link-list to all the menue items for ep2025.europython.eu/programme/* .

nikoshell:
Short links for pages. Menu have links tree structure, dead links and redirects are covered in astro config.
Close #1002


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-issue-1002.ep-preview.click|
|last update|2025-04-17T07:09:54.742Z|

astro.config.mjs
```diff
@@ -51,15 +51,15 @@ export default defineConfig({
...
-    "/cfp/": "/programme/cfp/",
     "/planning/": "https://forms.gle/riw6CvML8ck94A4V9",
     "/reviewers/": "https://forms.gle/4GTJjwZ1nHBGetM18",
-    "/rust-summit/": "/programme/rust-summit/",
+    "/programme/cfp/": "/cfp/",
+    "/programme/c-api-summit/": "/c-api-summit/",
+    "/programme/rust-summit/": "/rust-summit/",
+    "/programme/voting/": "/voting/",
+    "/programme/wasm-summit/": "/wasm-summit/",
+    "/sessions/": "/sessions/",
```
clytaemnestra: Isn't this supposed to be `/programme/sessions` -> `sessions`?

egeakman:
IIRC @cmaureir initially suggested keeping things under categories (folders). I think I like that structure a bit more than having everything in the root page directory.

clytaemnestra:
I like using folders too, since we might end up with many pages. The question is whether short URLs would still work with that structure? We need short URLs for promotional purposes -- `europython.eu/cfp` looks much better on slides and is easier to remember, than `europython.eu/programme/cfp`.

I think @nikoshell also mentioned short URLs for SEO reasons.

nikoshell:
> I like using folders too, since we might end up with many pages. The question is whether short URLs would still work with that structure? We need short URLs for promotional purposes -- `europython.eu/cfp` looks much better on slides and is easier to remember, than `europython.eu/programme/cfp`.
> 
> I think @nikoshell also mentioned short URLs for SEO reasons.

Ok, we can have both. Directory structure and short links for every page.
There will exists links like /programme/cfp and /cfp without redirections.

astro.config.mjs
nikoshell: We don't need redirects for pages, short and long links works by default.

clytaemnestra (commented):
@nikoshell I see long links on three pages - should these be updated and short ones used everywhere or it doesn't matter?

image

clytaemnestra (approved):

PR #1137, opened 2025-04-13
Add program & events overview 

clytaemnestra:
We have many different sessions and events across the page, but there's no single place where everything is mapped out. I wanted to create an overview where I can quickly scan what's available and click only on the pages I'm interested in, without opening each one individually.

Wdyt? I'm also open to ideas on how to make it more condensed and readable.


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-program-overview2.ep-preview.click|
|last update|2025-04-13T11:05:49.311Z|

egeakman (commented):
The content looks good, thank you Mia! Maybe we can add a link to the session types info that @hypha added in #989. Or maybe we can move that content here?

clytaemnestra:
I was checking the content and to me it feels too long. Ideally, it should be one sentence to keep it short and factual. I also think the main info belongs on the main page, not the FAQ. I agree it's best not to duplicate content. 

I'll have to think about it more, but I'm open to ideas on how to structure it.

PR #1165, opened 2025-04-19
Update preview endpoints.

nikoshell:


egeakman (approved):
Let's update it back to the 2025 URLs when the schedule is released on on Pretalx.

nikoshell:
> Let's update it back to the 2025 URLs when the schedule is released on on Pretalx.

Generally we should have preview env with mocked data and staging env which is prod like.

egeakman:
Also keep in mind that the 2024 schema is a bit different from 2025.

`gitx_url` needs to be changed to `gitx`
`bluesky_url` needs to be removed.

europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-fix-preview.ep-preview.click|
|last update|2025-04-19T20:48:15.989Z|

PR #1180, opened 2025-04-24
Update loadData function.

### LINKED ISSUE #1175: Builds randomly failing with `The collection "sessions" does not exist. Please ensure it is defined in your content config.`

nikoshell:
Fix #1175.
We add retries when we fetch data.
We checks that fetched json is valid and it is not empty.
Reduce calls for the same urls.


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-loaddata.ep-preview.click|
|last update|2025-04-26T18:29:57.795Z|

src/utils/dataLoader.ts
```diff
@@ -1,22 +1,47 @@
...
+        await new Promise((res) => setTimeout(res, retryDelay));
+      } else {
+        throw new Error(
+          `Failed to load data from ${apiUrl} after ${maxRetries} attempts`
+        );
+      }
+    }
   }
+
+  throw new Error(`Unexpected execution path`);
```
clytaemnestra: This line should never be reached, right? Is it common practice to include this kind of fallback?

nikoshell: Yes it is, and it's unexpected, but it's not obvious and editors complains, we can remove it.

src/utils/dataLoader.ts
```diff
@@ -1,22 +1,90 @@
-export async function loadData(apiUrl: any) {
+// Define a more specific response type or use generics
+type DataResponse = unknown;
+
+const dataCache: Record<string, DataResponse> = {};
```
clytaemnestra, line 4: When and how does `dataCache` get cleared or updated?

nikoshell: It's in memory cache used during build process, we have fresh every time we build.
Generally we had two requests to the same endpoint and we reduce it to one to each endpoint.

egeakman (approved):

PR #1181, opened 2025-04-25
Add infobox to schedule layout.

### LINKED ISSUE #1176: Note on the schedule page that the sessions might still change
Let's add the infobox at the top of the each of the schedule pages (Tutorials/Talks), that it's the early version of the schedule it might still change – just to manage the expectations if people plan their travel based on certain sessions.

https://ep2025.europython.eu/schedule/tutorials

https://ep2025.europython.eu/schedule/talks

nikoshell:
Close #1176.


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-infobox.ep-preview.click|
|last update|2025-04-26T17:50:05.098Z|

src/layouts/ScheduleLayout.astro
```diff
@@ -15,7 +15,13 @@ const { title, description } = Astro.props;
     <link href="https://fonts.googleapis.com/css2?family=Inter+Tight:ital,wght@0,100..900;1,100..900&display=block" rel="stylesheet">
   </Fragment>
   <>
+
+    <p class="bg-[rgb(255,252,204)] text-[rgb(153,101,21)] border-l-4 border-[rgb(255,204,0)] p-4 text-sm flex items-center space-x-2">
+      <i class="fas fa-exclamation-triangle"></i>
+      <span><strong>Important:</strong> This is an early version of the schedule. It might still change, so please manage your expectations when planning your travel based on specific sessions.</span>
```
clytaemnestra: let's make it a bit softer:

```suggestion
      <span><strong>Important:</strong> This is an early version of the schedule and may still change. If you're planning travel around specific sessions, please keep that in mind.</span>
```

egeakman:
Hey @nikoshell, thank you for adding this! Can we add this warning to sessions list page as well? And also maybe we could extract this to a component so it can be used everywhere when needed?

clytaemnestra (approved):

clytaemnestra:
@egeakman people mainly check the sessions page to get an idea of what kinds of talks will be there. They check the schedule when planning their trips, so the two are a bit different.

I think for the sessions page, instead of the dramatic exclamation mark, we can simply update the description and that should be enough. So, instead of:
> Here you can find a list of all the sessions at the conference.

I'd suggest something like:
> You can find the full list of sessions for the conference here. Please note that the list might still change.

I think this would look better and serve the purpose well.

egeakman:
> @egeakman people mainly check the sessions page to get an idea of what kinds of talks will be there. They check the schedule when planning their trips, so the two are a bit different.
> 
> 
> 
> I think for the sessions page, instead of the dramatic exclamation mark, we can simply update the description and that should be enough. So, instead of:
> 
> > Here you can find a list of all the sessions at the conference.
> 
> 
> 
> I'd suggest something like:
> 
> > You can find the full list of sessions for the conference here. Please note that the list might still change.
> 
> 
> 
> I think this would look better and serve the purpose well.

I agree. Though I think it would still be nice if we could extract this to a component, should be pretty trivial.

PR #1183, opened 2025-04-26
New Sections component and Layouts clean up.

### LINKED ISSUE #1139: Decrease side margins
When I open any page, the side margins are as wide as the text area itself, which makes the page unnecessarily long:

Image

As inspiration, we can take a look at JS Nation -- their content is centered, but the side margins are much narrower and looks much better:

Image

nikoshell:
Improvements related to issue #1139 which makes switch to wide layout simple.


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-layout-v2.ep-preview.click|
|last update|2025-05-05T10:23:17.025Z|

clytaemnestra:
I like the idea of having a wider main frame, but without headers and content centered, the page looks off and asymmetrical:

image

I don't think that simply centering the headers would fix the issue, though. We have text-heavy pages where centered headers might also look weird, for example:

image

---

I do think widening the main frame is an important feature to implement, but it probably needs a bit more design work and some rearrangement of other components to better fit the new layout.

nikoshell:
> I like the idea of having a wider main frame, but without headers and content centered, the page looks off and asymmetrical:

This can change when we will have sticky header, there will be better visual separation.

> I do think widening the main frame is an important feature to implement, but it probably needs a bit more design work and some rearrangement of other components to better fit the new layout.

We can stay with old width, all other changes are still valid and we can go with them if there is no new issues.
New section component and layouts clean up makes ease to change width for all pages.

src/components/schedule/day.astro
clytaemnestra, line 1: Sessions got a bit misaligned: 

image

clytaemnestra, line 1: Can we remove this title? It's redundant, too big, and we don't have titles on other pages. 

image

src/components/keynoters/keynoters.astro
clytaemnestra, line 1: I like how all sections are aligned now, but the Keynoters one has still bigger margin than the other ones. Is there any reasoning why to keep it like that or should we align it as well?

image

src/components/schedule/day.astro
nikoshell: Done

src/components/keynoters/keynoters.astro
nikoshell: Done

src/components/schedule/day.astro
nikoshell: Fixed.

clytaemnestra (commented):
Is this intentional or a mistake? I think the previous ones, without an anchor, looked better. Wdyt?

image

clytaemnestra (approved):
LGTM! Thanks.

PR #1227, opened 2025-05-14
Add sprints venue page

clytaemnestra:

PR #1232, opened 2025-05-15
Clean up

nikoshell:
Clean up in the code and components, content and design should not be affected.
In some place we replace old icons. For example hero section on main page.

We need some clean up before Tailwind 4 upgrade and adding Astro compress.
There will be part 2 of clean up just to split PRs.


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-cleanup.ep-preview.click|
|last update|2025-05-15T13:02:39.419Z|

clytaemnestra (approved):

src/components/Map.astro
```diff
@@ -1,10 +1,10 @@
...
     title="Map"
-    className="w-full aspect-video max-w-[800px] border-4 border-white rounded-lg shadow-lg"
-    allowFullScreen={true}
+    class="w-full aspect-video max-w-[800px] border-4 border-white rounded-lg shadow-lg"
+    allowFullScreen="true"
     loading="lazy"
     referrerPolicy="no-referrer-when-downgrade"
   ></iframe>
-);
+  </Fragment>
```
clytaemnestra, line 10: FYI I updated this in another PR: https://github.com/EuroPython/website/pull/1227/files#diff-d7d44fd3cc527b8dbcaa8b661a0eb769d94080a4ae4fccc535320ffdc9fa5fefR1

We need to have the component map generic and be able to pass coordinates from google maps, so it's reusable at multiple places

PR #1233, opened 2025-05-15
Add new Map component.

nikoshell:
Usage in the code: 
`<Map location="Prague Congress Centre" zoom={15} />`

More examples:
https://ep2025-map.ep-preview.click/test

### DIFF
--- a/src/components/Map.astro
+++ b/src/components/Map.astro
@@ -1,4 +1,6 @@
 ---
+import type { HTMLAttributeReferrerPolicy } from 'react';
+
 interface Props {
   location?: string;
   embedCode?: string;
@@ -8,15 +10,8 @@ interface Props {
   zoom?: number;
   mapType?: "roadmap" | "satellite" | "hybrid" | "terrain";
   title?: string;
-  width?: string;
-  maxWidth?: string;
-  aspectRatio?: string;
-  borderWidth?: string;
-  borderColor?: string;
-  rounded?: string;
-  shadow?: string;
   loading?: "eager" | "lazy";
-  referrerPolicy?: string;
+  referrerPolicy?: HTMLAttributeReferrerPolicy;
 }
 
 const {
@@ -64,17 +59,11 @@ else if (location) {
     <iframe
       src={src}
       title={title}
-      class="w-full aspect-video "
-      allowFullScreen={true}
+      class="w-full aspect-video"
+      allowfullscreen={true}
       loading={loading}
-      referrerPolicy={referrerPolicy}
+      referrerpolicy={referrerPolicy}
       style="border:0;"
     ></iframe>
   )}
 </div>
-
-<style>
-  .astro-map-container {
-    width: 100%;
-  }
-</style>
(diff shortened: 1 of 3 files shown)


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-map.ep-preview.click|
|last update|2025-05-15T19:49:47.785Z|

src/components/Map.astro
```diff
@@ -1,10 +1,80 @@
...
+
+if (embedCode) {
+  const srcMatch = embedCode.match(/src="([^"]+)"/);
+  src = srcMatch ? srcMatch[1] : "";
+}
+else if (lat && lng) {
+  src = `https://maps.google.com/maps?q=${lat},${lng}&z=${zoom}&t=${
+    mapType === "satellite" ? "k" :
+    mapType === "hybrid" ? "h" :
+    mapType === "terrain" ? "p" : "m"
```
clytaemnestra: Could you please explain this part? Not sure what `k`, `h`, `p` or `m` represents

src/components/Map.astro
```diff
@@ -1,10 +1,80 @@
...
+  }&output=embed`;
+}
+
+else if (mapId) {
+  src = `https://www.google.com/maps/embed?pb=!1m18!1m12!1m3!1d3000!2d-73.9857!3d40.7484!2m3!1f0!2f0!3f0!3m2!1i1024!2i768!4f13.1!3m3!1m2!1s${mapId}!2s!5e0!3m2!1sen!2sus!4v1684164357244!5m2!1sen!2sus`;
```
clytaemnestra: Can you please explain how does this work? If I enter a map id, for example `ChIJweACSJyUC0cRE4gGU06KHKY` from https://developers.google.com/maps/documentation/javascript/examples/places-placeid-finder, it returns me the coordinates of the Empire State Building. Am I using it wrong or it doesn't work?

image

src/components/Map.astro
```diff
@@ -1,10 +1,80 @@
...
+  zoom?: number;
+  mapType?: "roadmap" | "satellite" | "hybrid" | "terrain";
+  title?: string;
+  width?: string;
+  maxWidth?: string;
+  aspectRatio?: string;
+  borderWidth?: string;
+  borderColor?: string;
+  rounded?: string;
+  shadow?: string;
```
clytaemnestra: Could you please point where is this parameter used? Same for `maxWidth` and `width` - I don't see them below, but maybe I'm missing something

src/components/Map.astro
```diff
@@ -1,10 +1,80 @@
...
+      style="border:0;"
+    ></iframe>
+  )}
+</div>
+
+<style>
+  .astro-map-container {
+    width: 100%;
+  }
+</style>
```
clytaemnestra: What is the difference between `width: 100%` here and `w-full` in the class list on line 62? Do both need to be there?

src/components/Map.astro
nikoshell: You need to ask Google Maps team.

src/components/Map.astro
nikoshell: It works when you provide api key, which we dont' have yet.
I think that still it worth to have this case supported.

src/components/Map.astro
nikoshell: Those are left overs which should be removed.

clytaemnestra (approved):

PR #1241, opened 2025-05-17
Draft design of sponsor social card.

### LINKED ISSUE #1216: Generate sponsor cards programmatically
We generate speaker cards programmatically and we’d like to do the same for sponsor cards by using a template combined with sponsor logos.

Here’s an example from last year: https://x.com/europython/status/1810335569848770924, but this year we'd like to use standard dimensions 900x900. 

We don’t have a specific design yet, but I suggest using the speaker card as a base and making slight adjustments so it’s not the same, but still visually similar.

nikoshell:
Screen Shot 2025-05-17 at 17 05 25
Close #1216.


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-sponsor-social-card.ep-preview.click|
|last update|2025-05-21T07:39:07.068Z|

clytaemnestra (approved):

PR #1245, opened 2025-05-19
Possibility to add session to favorites.

nikoshell:
We can mark sessions as favorite by opening session page and clicking on the heart.
Favorite talks will be visible on the schedule.

Favorite talks are persist per device.

PC:
image
image

Mobile:
image


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-ilikeit.ep-preview.click|
|last update|2025-05-27T11:03:27.142Z|

nikoshell:
@egeakman @artcz @hypha Can you provide any feedback about this feature?

egeakman:
Hey! Thanks for this long awaited awesome PR! I'll try to have a look in a few days.

hypha:
Thank you so so much, Marcin, for the wonderful feature! A few preliminary thoughts: 

1. When favouring a talk and returning to the schedule, I need to manually refresh the page to see the favourite status reflected. Is it possible to update the state in the schedule view without requiring a refresh?

1. Would it be possible to add the ability to favourite talks directly from the schedule overview? Viewers can then immediately identify talks they're interested in without needing to open individual talk details.

1. The favouring functionality only works for talks and not tutorials. Is this intentional, or should we extend it to cover all session types?

1. Do you think it is useful to leave a note somewhere to to clarify that these favourites won't carry over to different sessions & devices? I’m happy to help draft or suggest wording if useful.

nikoshell:
> Thank you so so much, Marcin, for the wonderful feature! A few preliminary thoughts:
> 
>     1. When favouring a talk and returning to the schedule, I need to manually refresh the page to see the favourite status reflected. Is it possible to update the state in the schedule view without requiring a refresh?
> 
>     2. Would it be possible to add the ability to favourite talks directly from the schedule overview? Viewers can then immediately identify talks they're interested in without needing to open individual talk details.
> 
>     3. The favouring functionality only works for talks and not tutorials. Is this intentional, or should we extend it to cover all session types?
> 
>     4. Do you think it is useful to leave a note somewhere to to clarify that these favourites won't carry over to different sessions & devices? I’m happy to help draft or suggest wording if useful.

1. It is how cache works, I will check if we can do something with that. 
2. No.
3. It works for all sessions. What kind of session type is missed?
4. We try to collect a feedback to understand the issues. We will provide additional notes about this feature.

Thank you for all comments.

clytaemnestra:
@hypha it works for tutorials as well for me

image

hypha:
> It works for all sessions. What kind of session type is missed?
> @hypha it works for tutorials as well for me

@nikoshell @clytaemnestra, I was blind. Please ignore point 3. Apologies.

src/components/island/CodeHeart.svelte
```diff
@@ -0,0 +1,148 @@
+<script>
+  import { onMount } from 'svelte';
+
+  // Component props
+  export let mini = false;
+  export let codeId = "";
+  export let code = codeId;
```
clytaemnestra: why are there two `codeId` and `code`?

src/components/island/CodeHeart.svelte
```diff
@@ -0,0 +1,148 @@
...
+  <div class="code-heart">
+    <button
+      class="heart-button"
+      on:click={toggleFavorite}
+      aria-label={isFavorite ? "Remove from favorites" : "Add to favorites"}
+      title={isFavorite ? "Remove from favorites" : "Add to favorites"}
+    >
+      {#if isFavorite}
+        <svg xmlns="http://www.w3.org/2000/svg" class="heart-icon filled" viewBox="0 0 24 24" width="24" height="24">
+          <path fill="currentColor" d="M12 21.35l-1.45-1.32C5.4 15.36 2 12.28 2 8.5 2 5.42 4.42 3 7.5 3c1.74 0 3.41.81 4.5 2.09C13.09 3.81 14.76 3 16.5 3 19.58 3 22 5.42 22 8.5c0 3.78-3.4 6.86-8.55 11.54L12 21.35z"/>
```
clytaemnestra: the only difference between this one and the one on the line 92 is `fill="none"` vs `fill="currentColor"`. can this not be defined elsewhere as one icon and have `fill` passed where it's used?

clytaemnestra (approved):
overall LGTM - I just had two small questions

src/components/island/CodeHeart.svelte
nikoshell: We keep session code as index and we repeat it as array value.
That way we will be able to reuse local storage manager in other scenarios and we do it in the same way in all places.

src/components/island/CodeHeart.svelte
nikoshell: This part was refactored, we use fa icons.

egeakman:
I'll have some time to review this over the weekend. I don't want to hold this, but I would like to review it before merging it. So let's give it a few days if that's fine with everyone.

egeakman (commented):
This looks great, I really appreciate your work @nikoshell! One small thought: since we're not currently using Svelte elsewhere in the project, would you be open to rewriting this using Astro + vanilla JS or React? We've used React before, so the team might be more familiar with that stack, and sticking to Astro/JS would keep things simple. Introducing Svelte could require some extra knowledge + maintenance work for the team going forward. Totally understand if you had a reason for choosing it, I just wanted to bring it up in case a lighter option might work just as well here!

Also, I will try to implement these after June 3rd when I will have more availability:
- Sync favorites without refreshing the page across tabs
- Ability to favorite a session from the main schedule page (as @hypha mentioned)

PR #1262, opened 2025-05-23
Upgrade to tailwind4

### LINKED ISSUE #1038: Migrate tailwindcss to v4, and use vite it through vite plugin

nikoshell:
Close #1038


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-tailwind4.ep-preview.click|
|last update|2025-05-26T19:59:10.442Z|

egeakman (commented):
Let's do the migration after the conference. This version changed how tailwind handles the priority of the passed classes. So it's really hard for us to have a clear idea of what still works and what doesn't given the amount of content we have.

nikoshell:
> Let's do the migration after the conference. This version changed how tailwind handles the priority of the passed classes. So it's really hard for us to have a clear idea of what still works and what doesn't given the amount of content we have.

Multiple cleanups was done to make the migration. 
So this PR do exactly that, it is part of unifying content, styles and colors, and next PR will go the same direction.
I don't see any reason to not merge it if there is no real bugs reported.

We also don't have huge amount of content, it's ~30 pages + ~3 template, most of the content is markdown styled with prose and typography plugin directly.

For more features we need to go with this
image

clytaemnestra:
> Let's do the migration after the conference. This version changed how tailwind handles the priority of the passed classes. So it's really hard for us to have a clear idea of what still works and what doesn't given the amount of content we have.

I totally get the idea of not wanting to break things right before the conference, especially with the traffic we now have. But we’re in a really good spot right now: we’re merging stuff quickly, getting reviews fast, and things are moving. That’s why I’d actually not wait until after the conference.

In my experience, once the conference is over, everyone kind of switches to vacation mode, and it gets way harder to find someone who has the time (or energy) to jump in and fix things. Right now, if something goes wrong, we can catch and fix it fast and that’s a big plus.

 I can set aside half an hour this weekend to click through all the pages and make sure we catch all bugs.

egeakman:
If you believe we can keep the effects contained, let's go for it!

Thanks for taking the work of checking the pages @clytaemnestra!

clytaemnestra (commented):
@nikoshell I went through all the pages and created a ticket summarizing the issues I found: https://github.com/EuroPython/website/issues/1264. Nothing major, all smaller issues. 

If you’d prefer to address them in a separate PR, let me know; I can approve this one, and we can handle the rest later. Or if you’d rather include the fixes here, just ping me once it’s ready for another review. I only checked the UI for now; I haven’t checked the changelog and looked at the code yet, but I can go through it once everything’s finalized.

clytaemnestra (approved):
LGTM!

PR #1278, opened 2025-05-28
Update keynoters.

nikoshell:
image


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-update-keynoters.ep-preview.click|
|last update|2025-05-30T08:00:52.833Z|

egeakman:
Is there a specific reason why we redirect to YouTube instead of embedding the player like we did before?

nikoshell:
> Is there a specific reason why we redirect to YouTube instead of embedding the player like we did before?

So, other way around, we just embedding the player instead of redirect to YouTube.

clytaemnestra:
> > Is there a specific reason why we redirect to YouTube instead of embedding the player like we did before?
> 
> So, other way around, we just embedding the player instead of redirect to YouTube.

@nikoshell if I click on the 'play' button, it redirects me to YT and I can't play it embedded. currently using Chrome version 136.0.7103.116. 

I think it would be better if there would be no redirect to YT, unless there are technical reasons not to play it embedded directly on the website. 

image

nikoshell:
> Is there a specific reason why we redirect to YouTube instead of embedding the player like we did before?

Sorry, it looks that you ware right I didn't understand your comment.

nikoshell:
> > > Is there a specific reason why we redirect to YouTube instead of embedding the player like we did before?
> > 
> > 
> > So, other way around, we just embedding the player instead of redirect to YouTube.
> 
> @nikoshell if I click on the 'play' button, it redirects me to YT and I can't play it embedded. currently using Chrome version 136.0.7103.116.
> 
> I think it would be better if there would be no redirect to YT, unless there are technical reasons not to play it embedded directly on the website.
> image

There was a bug. 
Generally whole idea was to have YouTube player for all markdown content not only for md pages.
fyi, right now we use two different renderers

nikoshell:
YouTube feature for markdown sections have some issue across dev and prod I will prepare it as separate PR.

egeakman (approved):
Thanks, let's do the YouTube embedding in another PR!

PR #1291, opened 2025-06-05
YouTube player for markdown

nikoshell:
New YouTube component which can be also used in any markdown content.
YT links are replaced with player.


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-yt.ep-preview.click|
|last update|2025-06-09T07:08:52.481Z|

clytaemnestra (commented):
~~Could you please describe the problem this PR is addressing?~~

Never mind, I got it.

astro.config.mjs
```diff
@@ -114,23 +116,27 @@ export default defineConfig({
...
+            CSS: false,
+            SVG: false,
+          }),
+          dontDie(),
+        ]),
   ],
   output: "static",
   build: {
-    minify: true,
+    ...(fastBuild ? {} : { minify: true }),
```
clytaemnestra, line 147: could you please point when and where is `fastBuild` set to `true`?

src/components/ui/YouTube.astro
```diff
@@ -1,21 +1,78 @@
...
-} = Astro.props;
-
-const attrId = attrs.id || '';
+	...attrs
+} = Astro.props as Props;
 
 const defaultClass = 'border-4 border-white rounded-lg shadow-lg';
 const className = `${defaultClass} ${userClass}`.trim();
+
+const idRegExp = /^[A-Za-z0-9-_]+$/;
```
clytaemnestra: according to the documentation it's 11 only 

```suggestion
const idRegExp = /^[A-Za-z0-9-_]{11}$/;
```

src/components/ui/YouTube.astro
```diff
@@ -1,21 +1,78 @@
...
+}
+
+const videoid = extractID(id);
+const posterFile =
+	{
+		max: 'maxresdefault',
+		high: 'sddefault',
+		default: 'hqdefault',
+		low: 'default',
+	}[posterQuality] || 'hqdefault';
```
clytaemnestra, line 48: isn't this supposed to be values as per line 17 above?

```suggestion
	}[posterQuality] || 'default';
```

astro.config.mjs
nikoshell: There is no place where it's set to true as this is development flag use do speed builds by building only base.

src/components/ui/YouTube.astro
nikoshell: https://github.com/justinribeiro/lite-youtube?tab=readme-ov-file#attributes
It's correct.

clytaemnestra (approved):

egeakman (approved):
Thank you Marcin!! One small question below:

astro.config.mjs
```diff
@@ -27,6 +27,8 @@ if (!gitVersion) {
   }
 }
 
+const fastBuild = false;
```
egeakman: Can we make this passable like `--mode`? It's fine to add it in another PR.

nikoshell: mode is astro argument, technically we can override it but there is no reason for that
I can do it as new flag.

egeakman: Yeah yeah, I meant as a new flag, similar to how we implemented --mode.

nikoshell: I add it already as env variable EP_FAST_BUILD and set to `true` in dotenv for development.

PR #1297, opened 2025-06-06
add arm's logo

hypha:
FYI, the display.png image might be slightly short, but I just used the same banner they used everywhere, and I think with the small white border, it looks OK. Plus, hopefully, we can get their own submission if this is to be included.


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-sponsor-logo-arm.ep-preview.click|
|last update|2025-06-07T21:15:37.313Z|

src/content/sponsors/arm/index.md
```diff
@@ -0,0 +1,25 @@
+---
+name: Arm
+url: http://www.arm.com
+description:
+  Born more than 30 years ago with the goal of designing a computer intended to
+  run on a battery, Arm has become a global compute platform delivering advanced
+  solutions that allow the world’s leading technology companies to unleash the
+  unprecedented experiences and capabilities of AI.
```
clytaemnestra, line 8: The text is okay, but the only way how to find it is by using search. I think we should change the links of logos, so if I'm on the homepage and I click on the logo, it leads me directly to the sponsor page. It's out of scope of this PR, though. 

Cc @nikoshell

clytaemnestra (approved):

src/content/sponsors/arm/index.md
hypha: I think they are not supposed to be shown / visible yet because we haven't made decisions on how to present these dedicated pages. I forgot to make this a draft, now done in https://github.com/EuroPython/website/pull/1297/commits/866c2f2c63632b8631089df8ecaa025cf9b7e828

nikoshell (approved):

PR #1314, opened 2025-06-11
Add animation

nikoshell:


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-hero3.ep-preview.click|
|last update|2025-06-16T18:08:34.472Z|

egeakman:
I personally like it, but this might raise an accessibility issue. There should be an easy-to-reach option to disable it IMO.

cc @ADubhlaoich @cybit

nikoshell:
> I personally like it, but this might raise an accessibility issue. There should be an easy-to-reach option to disable it IMO.
> 
> cc @ADubhlaoich @cybit

This should not be an issue we will respect users browser settings "prefers-reduced-motion"
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

clytaemnestra:
I'm not a big fan of it, but if the majority prefers this option, I can live with it

nikoshell:
I we want more animation on the website we can improve whole layout, this PR is just idea to collect some feedback.

PR #1352, opened 2025-06-25
Add jobs order by the tier

nikoshell:


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-jobs-order.ep-preview.click|
|last update|2025-06-25T17:00:33.069Z|

src/pages/jobs.astro
```diff
@@ -46,6 +55,20 @@ sponsors.forEach((sponsor) => {
...
+  .sort((a, b) => {
+    const aTierIndex = tiers.indexOf(a.data.tier as typeof tiers[number]);
+    const bTierIndex = tiers.indexOf(b.data.tier as typeof tiers[number]);
+
+    // If tier not found in array, put it at the end
+    const aIndex = aTierIndex === -1 ? tiers.length : aTierIndex;
+    const bIndex = bTierIndex === -1 ? tiers.length : bTierIndex;
+
+    return aIndex - bIndex;
+  });
```
clytaemnestra, line 71: Not a JS developer, but this looks complicated to me - I think we can assume that each sponsor has a tier assigned. Otherwise we wouldn't be able to create one, no? 

In that case, I think something as simple as `sponsors.filter((s) => s.data.jobs.length).sort((a, b) => tiers.indexOf(a.data.tier) - tiers.indexOf(b.data.tier));` might work (I didn't try to run it, I'm guessing here) and is more readable, no?

clytaemnestra (approved):

src/pages/jobs.astro
nikoshell: We care only about sponsors with tier which allows to job posts. Array of tiers is shorter then all tiers that we have.

PR #1366, opened 2025-06-30
EP2025 sponsor logo anaconda

hypha:
-  @artcz previously requested us to list these main add-ons to the website, so I added Open Space and Sprints tiers. I put them before and after finaid so the same logo doesn't appear consecutively which might be a big ugly. 
- Previously, sponsors could only belong to one tier. This is a poor attempt to allow sponsors like Anaconda to be listed for multiple tiers, such as "Bronze, Open Space, Sprint" and appear in all three sections in the sponsors page. There is probably way better ways to do this.


src/components/sections/sponsors/sponsors.astro
```diff
@@ -17,8 +17,10 @@ const tiers = [
   "Startups",
   "Bronze",
   "Patron",
+  "Sprint",
   "Financial Aid",
-  "Supporters",
+  "Open Space",
```
clytaemnestra, line 22: Maybe to have one category `Events Sponsors` or `Activities Sponsors` and inside of it to have `Sprints` and `Open Space` and under both to have their logo? Wdyt? 

Also cc @nikoshell for opinions.

europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-sponsor-logo-anaconda.ep-preview.click|
|last update|2025-06-30T14:30:49.963Z|

src/components/sections/sponsors/sponsors.astro
hypha: I think we explicitly agreed with sponsors to be listed as an open space or sprint sponsor, and that was also the brief I got, so though I liked the idea aesthetically, it might not work out this year.

hypha: sorry! I completely misread your message. So you want a nested tiers? That would work too but not sure how it should look like. Happy for Marcin to take over.

clytaemnestra: okay! that's okay for me then :)

clytaemnestra: yes, I meant nested tiers like this
image

nikoshell: > Maybe to have one category `Events Sponsors` or `Activities Sponsors` and inside of it to have `Sprints` and `Open Space` and under both to have their logo? Wdyt?
> 
> Also cc @nikoshell for opinions.

From my perspective Open Space and Sprints are not tiers, both also have or will have dedicated page where we can mention sponsors with their logos.

I also don't remember any event that duplicate logos in sponsors section.

nikoshell:
(@hypha lets use drafts as designed and leave red labels for critical things)

src/components/sections/sponsors/sponsors.astro
clytaemnestra: I like the idea of having it only on respective pages! That's how we had Kraken last year for PyLadies lunch. I'm not sure if there wasn't any other deal with Anaconda, so maybe this is a question for the requestor of the feature-- @artcz if you're okay with having it on pages only?

hypha:
closing in preference for better implementation in https://github.com/EuroPython/website/pull/1372

PR #1383, opened 2025-07-02
Sprints page with filter

nikoshell:
image


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-sprints.ep-preview.click|
|last update|2025-07-09T18:02:20.128Z|

rodrigogiraoserrao (approved):
This looks good to me in principle but I'd love if someone could reassure me w.r.t. the fact that the preview, when open on the sprints weekend page, only shows the sprints cards and doesn't show the other info that's already live on the website.

rodrigogiraoserrao:
(Also, I've added stubs for the CPython & BeeWare project sprints and I'll reach out to the folks I talked to initially to get their email addresses/full descriptions for their project cards.)

src/content/sprints/cpython.md
```diff
@@ -0,0 +1,17 @@
...
+contactPerson:
+  name: "Petr Viktorin"
+  email: "name@example.com"
+links:
+  - title: "Code of Conduct"
+    url: "https://www.python.org/psf/conduct/"
+  - title: "CPython's Developer Guide"
+    url: "https://devguide.python.org/"
+  - title: "Easy Issues"
+    url: "https://github.com/python/cpython/issues?q=is%3Aissue+is%3Aopen+label%3Aeasy"
```
clytaemnestra: I'd put guide on the first place, then easy issues, then CoC

clytaemnestra (approved):

nikoshell:
> This looks good to me in principle but I'd love if someone could reassure me w.r.t. the fact that the preview, when open on the sprints weekend page, only shows the sprints cards and doesn't show the other info that's already live on the website.

Correct this is the goal, I will update PR to have correct preview view.

src/content/sprints/cpython.md
nikoshell: Ok

egeakman (approved):
Thanks Marcin! Looks good to merge.

Let's add a notice on how to add a new project on this page in the next PR.

src/content/sprints/cpython.md
rodrigogiraoserrao: (Just for reference, the order of these 3 links was copied from the 2024 sprint project.)

rodrigogiraoserrao (commented):
Why are there two “sprints template” files, the `_sprints_template.md` and the `sprints_template.md`?

Also, in `_sprints_template.md`, why do we instruct folks who make the PRs to keep `draft: true` in the project they're adding?
If they're adding it, it's not supposed to be a draft; it's supposed to be live.
(Maybe I misunderstood the point of `draft: true`.)

nikoshell:
> Why are there two “sprints template” files, the `_sprints_template.md` and the `sprints_template.md`?
> 
> Also, in `_sprints_template.md`, why do we instruct folks who make the PRs to keep `draft: true` in the project they're adding? If they're adding it, it's not supposed to be a draft; it's supposed to be live. (Maybe I misunderstood the point of `draft: true`.)

We use draft flag in multiple places to manage content, it's useful in case when we want commit but not publish content, when it  is finial we need to remove it. Content with draft: true is visible with watermark on not prod environments including preview environment.

 _sprints_template.md is a template and we need to prepare description how to use it.
  sprints_template.md renamed to example_draft.md was created to show how draft looks on preview and it's not visible on prod

PR #1432, opened 2025-07-12
Change background color

clytaemnestra:
In dark mode the dark navbar is not that visible and looks like it's part of the browser. I changed the color to orange, so it's visible in both dark and light modes.

### DIFF
--- a/src/components/Header.astro
+++ b/src/components/Header.astro
@@ -11,7 +11,7 @@ import links from "@data/links.json";
 <section
   id="navbar"
 >
-<a href="/remote" class="hover:underline hover:text-primary-hover text-center px-2 bg-secondary text-white font-bold fixed top-0 w-full h-[40px] z-50 flex items-center justify-center">Can’t join us in Prague? Attend remotely — online options available!</a>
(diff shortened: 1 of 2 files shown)


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-navbar-bg-color.ep-preview.click|
|last update|2025-07-13T23:00:13.819Z|

hugovk:
There's a dark mode?! How do I enable it?

The contrast ratio when hovering the mouse over it is very low, and it's hard to read:

image

Can we change it? Perhaps keep it white, we have the underline to indicate hover.

nikoshell:
> There's a dark mode?! How do I enable it?
> 
> The contrast ratio when hovering the mouse over it is very low, and it's hard to read:
> image
> 
> Can we change it? Perhaps keep it white, we have the underline to indicate hover.

Thank you we also notice that.
I was thinking about something from current palette colour
image

clytaemnestra:
@hugovk no dark mode yet, but hopefully for 2026! By dark mode I meant dark mode of the browser:
image

---

Agreed with @nikoshell to keep the orange, but let's have white as text color and the same white+underline on hover. That way the contrast should be okay.

nikoshell (approved):

PR #1434, opened 2025-07-12
Update prefetch settings

nikoshell:
There is no reason to use prefech when we don't use any html files caching.


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-cache-update.ep-preview.click|
|last update|2025-07-13T20:47:49.068Z|

clytaemnestra (approved):

PR #1445, opened 2025-07-15
Fix broken link to Prague USEIT Map

aleszoulek:
`https://www.use-it.travel/cities/detail/prague/` seems to be broken, the correct working link is `https://www.use-it.travel/cities/prague/` 

🖼️ Preview available 🖼️ : https://ep-website--1445.org.readthedocs.build/


hugovk (approved):
Thank you!

PR #1446, opened 2025-07-15
We will remove old builds from now

nikoshell:


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2025-space5.ep-preview.click|
|last update|2025-07-15T08:56:21.096Z|

egeakman (dismissed):

Makefile
```diff
@@ -67,7 +67,7 @@ ls -1 */ \
   | grep \"^2025\" \
```
egeakman: Why do we only match directories starting with 2025?

egeakman: @nikoshell

nikoshell:
We will back to this but as far we can close it.

PR #1468, opened 2025-07-18
Expand scope of Apache Arrow sprint

rok:
There seems to be interest, so we'd like to propose a slightly different sprint.

🖼️ Preview available 🖼️ : https://ep-website--1468.org.readthedocs.build/


hugovk (approved):
Thank you!

PR #1472, opened 2025-07-19
Remove service worker registration

patrick91:
This also needs a `Clear-Site-Data` which @artcz is working on

### DIFF
--- a/src/layouts/Layout.astro
+++ b/src/layouts/Layout.astro
@@ -110,16 +110,6 @@ footer {
 </style>
 
 <script>
-if ("serviceWorker" in navigator) {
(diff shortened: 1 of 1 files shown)


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://remove-service-worker.ep-preview.click|
|last update|2025-07-19T13:04:28.929Z|

patrick91:
CleanShot 2025-07-19 at 15 08 12@2x

tested here and doesn't try to add a service worker 😊

So once we add the header, things should be fixed 😊

artcz (approved):

PR #1529, opened 2026-03-09
bugfix: sponsor packages page now has the same width as other pages

anorthall:
Previously, the sponsor packages page (`/sponsorship/sponsor/`) was not the same width as all other content pages (it was narrower and looked odd when navigating between pages).

This was due to `MarkdownLayout` having different default styling than `SectionLayout`, the latter of which is used for most pages.

I have fixed it by adding a `full` parameter to the frontmatter which is then handled in `MarkdownLayout` via `[...slug].astro` to allow pages to be full width. The `sponsor.mdx` page was updated with `full: true`.

# Screenshots

## Before

CleanShot 2026-03-09 at 05 18 16@2x

## After

CleanShot 2026-03-09 at 05 18 34@2x

# Mobile

Designed / layout is unchanged on mobile.

🖼️ Preview available 🖼️ : https://ep-website--1529.org.readthedocs.build/

### DIFF
--- a/src/content/config.ts
+++ b/src/content/config.ts
@@ -11,6 +11,7 @@ const pages = defineCollection({
     title: z.string(),
     subtitle: z.string(),
     toc: z.boolean().optional().default(true),
+    full: z.boolean().optional().default(false),
   }),
 });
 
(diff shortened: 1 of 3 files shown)


clytaemnestra:
@anorthall  LGTM, thanks for the PR! 🙂

PR #1530, opened 2026-03-10
Revert "bugfix: sponsor packages page now has the same width as other pages"

clytaemnestra:
Reverts EuroPython/website#1529 

I realized, after merging, that the narrow width was intentional - it's the same design as for the tickets page, so let's have it rather narrow, since the content was designed around it. 
Discussed with Andrew in a DM.


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://revert-1529-an-fix-sponsor-page-width.ep-preview.click|
|last update|2026-03-10T13:03:24.438Z|

PR #1592, opened 2026-05-26
update available sponsor slots

hypha:
Simplified the way to handle slotLeft a little.
- totalSlots is now an integer (where applicable, not used for Patron and Bronze)
- slotsLeft behaves as follows:
  - If it is a number, it renders as “X out of Y slots available”
  - If it is a string, it renders directly as provided (e.g. "Exclusive", "Fully booked", "Yearning for Dogs", etc)


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2026-sponsor-slots-update.ep-preview.click|
|last update|2026-05-26T13:50:19.381Z|

deutranium (approved):
merging with a very big thanks <3

PR #1643, opened 2026-06-18
Fix timezone bug for local dev

hugovk:
I'm in Helsinki, east of UTC.

Presumably the server or GitHub Actions build is using UTC for the prod build.

Using local midnight time means the day calculation is wrong. Use noon UTC instead.

# Prod

image

# Local dev before

image

# Local dev after

image

### DIFF
--- a/src/components/schedule/day.astro
+++ b/src/components/schedule/day.astro
@@ -349,7 +349,7 @@ posters.forEach((poster) => {
 });
 
 
-const date = parseISO(dayName);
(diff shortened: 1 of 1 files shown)


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://fix-timezone-bug.ep-preview.click|
|last update|2026-06-18T11:10:06.201Z|

nikoshell (approved):

PR #1689, opened 2026-06-26
Add childcare info

emmaco:
add childcare page and link to childcare page from a11y page

🖼️ Preview available 🖼️ : https://ep-website--1689.org.readthedocs.build/


hypha (commented):
Great work @emmaco! I wonder if it makes sense to put the childcare page on the menu as well, perhaps under Attend, so it's easier to find?

nikoshell (approved):

src/data/nav.ts
```diff
@@ -65,6 +65,7 @@ const L = {
     external: true,
   },
   accessibility: { label: "Accessibility", url: "/accessibility" },
+  childcare: { label: "Childcare", url: "/childcare" },
```
hypha, line 68: I don't see childcare on the menu from preview. Looks like in this brave new way of adding to the menu, this part just defines the registry, and to actually reference it, perhaps it needs to be added after https://github.com/emmaco/website/blob/a4ff75fae0f05d93193a0d680105963d5e8121f3/src/data/nav.ts#L176

emmaco: Ah, I think I just missed adding it there. I'll update now.

hypha (approved):
Thank you Emma! I am so happy to see it on the menu now <3

emmaco:
If it's okay, are you able to merge please @hypha? I don't think I have permissions for that.

hypha:
More than OK @emmaco! Sorry I thought perhaps you preferred to merge when ready (and that you had permission). Merged now with big thanks!

PR #1690, opened 2026-06-26
add participation encouragement for grant recipients

hypha:


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2026-finaid-warmth.ep-preview.click|
|last update|2026-06-26T11:12:14.379Z|

thepetk (approved):
lgtm! We could merge it as is.

However the question might still remain: Folks might need to identify which is event is free and which needs a ticket maybe. So we could also explain that the ticket granted via Finaid gives access to conference and sprints while for all other events they need to register (if registration applies) ? WDYT?

hypha:
@thepetk yeah it's difficult because we do not want to lay out policy again in the message that is meant to be encouraging. I added some gentle clarification here: https://github.com/EuroPython/website/pull/1690/changes/74fa2a33fa4508879a454050172d11e37d733f1a WDYT?

thepetk:
> @thepetk yeah it's difficult because we do not want to lay out policy again in the message that is meant to be encouraging. I added some gentle clarification here: [74fa2a3](https://github.com/EuroPython/website/commit/74fa2a33fa4508879a454050172d11e37d733f1a) WDYT?

looks good to me!

PR #1692, opened 2026-06-26
Talks Tutorials grid and Sessions at the same time

nikoshell:


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2026-talks-tutorials.ep-preview.click|
|last update|2026-07-10T20:13:11.261Z|

dariagrudzien:
1. Open spaces tab is empty, something should be there, at least some placeholder text in the middle of the page? So it doesn't look like it's broken, which it now does.
2. Nitpick: Talks vs Tutorials swap is not intuitive to me. Would it make sense to see both?

dariagrudzien (dismissed):
Open Spaces page looks broken, can we put some one-liner in the middle of the page that details are coming soon?

hypha:
I like the posters page. But I don't think the schedule works well
- When I click `Schedule`, before the conference I’d expect it to default to **Monday**, and during the conference to default to **today**. From there, it should be easy to navigate between **Mon–Fri**. Right now it goes straight to `/talks`, and it takes me longer than it should to realise how to navigate to tutorials.

- For the tutorial days, Fishbowl and multifunctional rooms are for summits, so they should come _after_ S4B in the order.

- During the talk days, the events outside the main 5 tracks should either live somewhere else (for example like in #1695, with maybe a small note at the top signposting to them for visibility), or at least come *after* the main tracks rather than before. If we keep them in the same schedule view, I think the display needs rethinking. Right now the 5 talk-track columns become very narrow and long especially for Friday, which makes it hard to scan, follow, and tell where things begin and end. (And since this is hard to do, I vote for keeping only the 5 main tracks here).

* Also, the posters on Friday look pretty broken right now

image

nikoshell:
> I like the posters page. But I don't think the schedule works well
> 
>     * When I click `Schedule`, before the conference I’d expect it to default to **Monday**, and during the conference to default to **today**. From there, it should be easy to navigate between **Mon–Fri**. Right now it goes straight to `/talks`, and it takes me longer than it should to realise how to navigate to tutorials.
> 
>     * For the tutorial days, Fishbowl and multifunctional rooms are for summits, so they should come _after_ S4B in the order.
> 
>     * During the talk days, the events outside the main 5 tracks should either live somewhere else (for example like in [add side events page #1695](https://github.com/EuroPython/website/pull/1695), with maybe a small note at the top signposting to them for visibility), or at least come _after_ the main tracks rather than before. If we keep them in the same schedule view, I think the display needs rethinking. Right now the 5 talk-track columns become very narrow and long especially for Friday, which makes it hard to scan, follow, and tell where things begin and end. (And since this is hard to do, I vote for keeping only the 5 main tracks here).
> 
> 
>     * Also, the posters on Friday look pretty broken right now
> 
> 
> image

As I understand those are already covered by other PRs.

gbdlin (approved):

PR #1695, opened 2026-06-27
add side events page

dariagrudzien:
Based on feedback from attendees from the previous year, adding a dedicated page to keep all sessions and events other than talks & tutorials on a single page, so they are easier to discover - and find at the venue.

image


src/content/events/language-summit.md
```diff
@@ -0,0 +1,13 @@
...
+title: Python Language Summit
+date: "2026-07-14"
+time_start: "09:00"
+time_end: "17:00"
+location: "ICE Kraków Congress Centre, room TBA"
+url: /language-summit
+description:
+  CPython and alternative implementation developers gather to share information,
+  discuss common problems, and solve them. Invite-only for Python core
+  developers.
```
hugovk: Invite-only, but not only the Python core team.

The long version is:

> We welcome Python core team and triage team members, active contributors to CPython and alternative Python implementations, and other community members with a topic to discuss with the core team.

https://ep2026.europython.eu/language-summit/#who-can-attend

Maybe just say "invite only"?

```suggestion
  discuss common problems, and solve them. Invite only.
```

hugovk (commented):
Great idea!

europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://daria-alt-schedule.ep-preview.click|
|last update|2026-06-28T20:27:08.558Z|

nikoshell:
Cons:
- We are repeating that what we have in the menu but with item "Side Events"
image

- We create schedule/events/  new page that have timestamps and events but with new design 

Pros:
- Simple information about side events in one place

Suggestions:
- we need to organize menu in better way 
- we can use current schedule theme to present data
image
- those boxes can be used to organize schedule sub navigation
- other option is to use this layout
image

Things to avoid:
- css with fonts manipulations, at least typography and colors we try to have common.

dariagrudzien:
@nikoshell Thanks for the comment!

1. Duplicating pages - this is by design. Last year we had all events in the navbar, and people still struggled to find them 🤷‍♀️ So this is meant as an additional way to see the same information, in the hope that it will help attendees find events they would otherwise miss. I also think that seeing them all in a single view, with all time and room details is a better way to present the offer, then just let people read the details on specific event page.

2. `new page that have timestamps and events but with new design` - that's also my continuous decision to display the information differently. My initial instinct was to duplicate the schedule format, but in my experience that's more confusing - you have 2 different tabs open, they mostly look the same, but they have different information. I decided I'd rather make this alternative events page visually very distinct, consistent with the rest of the website but in complete contrast to the schedule. 

An additional reason not to pick the schedule format is this: talks and tutorials fit into the same time slots, while all other events have very different timing. Some are full day, some are 1 hour, some will be 2-3 hours, and overall it's not going to be 100 sessions which have to fit into a page, but maybe 10-20. In my opinion that requires more flexibility. I'm open to alternatives, but I don't think that duplicating the talks schedule is the way to go. Unless we want to try and add all those side events into the schedule, which is another thing which I considered, but I think it would be overwhelming and would break the visual integrity of the current schedule. Maybe @cmaureir has an opinion about this?

3. `we need to organize menu in better way` - I don't know, do we? It looks quite good to me, particularly the Programme section.

4. Using the Poster layout - I didn't know that we have it, thanks for pointing it out. I'm open to trying it out to see how it will work. What's really missing from that layout for me is a quick visual info about the time and location - you need to click on the details of the specific event. It also has less space to add those 2-3 sentences about the event which explain what it's about. I do think it's a better format than just duplicating the schedule, but I think that making those pages visually similar is not good for the users. I think that keeping the layouts different allow returning users to quickly identify if they are on the correct page. It it my personal, uneducated opinion though, I'd appreciate someone with UX experience weighing in :)

hypha:
@dariagrudzien I really like the idea as well. Happy to do a review of the content too, but at the moment I can see some locations are still marked as TBD even though they’re already available. Not sure if those are just placeholders for now in draft stage or ready for review.  A few initial thoughts:

- I think it would be really beneficial to put it in the menu. At the moment, even though you’ve added it in `nav.ts` `  sideEvents: { label: "Side Events", url: "/schedule/events" },` because it’s nested I don’t actually see it in the menu.

- I’m a little unsure about the term “side event”, as it might suggest they’re secondary to the talks, which may or may not be true depending on the person. I’m struggling to think of a better term, so here’s a slightly wonky one: Beyond the Talks in the menu, and Events Beyond the Talks as the page title.

- In fact, during the conference it might even be nice to link the page as a CTA alongside the talk schedule on the homepage

clytaemnestra:
Thanks for the PR! I like the idea too! 

+1 on adding it to the menu, I couldn't find the page. 

I think duplicating pages is okay as long as it serves a different purpose, for example detailed information vs overview etc. There's also this: https://ep2026.europython.eu/overview/

I like the poster layout, but it's missing times, as @dariagrudzien pointed out. So if we're not able to take that box and add times, I'd rather leave the layout that is in this PR.

dariagrudzien:
After a discussion with @nikoshell we agreed on a different direction, but we'll make sure that side events are easy to find.

PR #1699, opened 2026-06-29
Add #EuroPython #Python hashtags to bingo share

hugovk:
Also list social networks in same order as homepage:

image


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://share-bingo.ep-preview.click|
|last update|2026-07-03T23:50:55.949Z|

src/components/island/BingoCard.svelte
```diff
@@ -53,10 +53,10 @@
     const count = attended.length;
     const base = `I've attended ${count} EuroPython conference${count !== 1 ? 's' : ''}! 🐍`;
     const body = count > 0 ? `${base}\n\n${attended.join(' · ')}` : base;
-    return `${body}\n\nWhat about you: ${BINGO_PAGE_URL}`;
+    return `${body}\n\nWhat about you? ${BINGO_PAGE_URL}\n#EuroPython #Python`;
   }
 
-  const BINGO_PAGE_URL = 'https://ep2026.europython.eu/#bingo';
+  const BINGO_PAGE_URL = 'https://ep2026.europython.eu/bingo';
```
dariagrudzien, line 59: This was suggested to me before, but other sections have URLS like `https://ep2026.europython.eu/#krakow` - don't we want to be consistent? And does the hash even matter?

src/components/island/BingoCard.svelte
```diff
@@ -53,10 +53,10 @@
     const count = attended.length;
     const base = `I've attended ${count} EuroPython conference${count !== 1 ? 's' : ''}! 🐍`;
     const body = count > 0 ? `${base}\n\n${attended.join(' · ')}` : base;
-    return `${body}\n\nWhat about you: ${BINGO_PAGE_URL}`;
+    return `${body}\n\nWhat about you? ${BINGO_PAGE_URL}\n#EuroPython #Python`;
```
dariagrudzien:
```suggestion
    return `${body}\n\nWhat about you? ${BINGO_PAGE_URL}\n\n#EuroPython #Python`;
```

dariagrudzien: This would help readability

dariagrudzien:
@hugovk I don't really know why, but this change breaks LinkedIn sharing by removing the URL, by clipping everything after the last question including the URL:

this PR
image

prod
image

dariagrudzien (dismissed):

src/components/island/BingoCard.svelte
hugovk: I don't think we need them, they look a bit weird, and when I click around the site, I don't see any hashes like this in URLs (not counting mid-page anchors)

hugovk:
> I don't really know why, but this change breaks LinkedIn sharing by removing the URL, by clipping everything after the last question including the URL:

Hmm, this is what Claude suggests:

>   Root cause: The ? introduced in commit c191e311 ("What about you:" → "What about you?").
>   The text is sent correctly encoded as %3F, but LinkedIn's shareActive flow decodes the
>   text parameter and re-parses it internally, so a literal ? in the decoded text truncates
>   everything after it. The # characters are not the issue — prod's URL contained /#bingo
>   and shared intact, so the new hashtags can stay.
> 
>   The fix in src/components/island/BingoCard.svelte:51: buildShareText() now takes a
>   forLinkedIn flag. LinkedIn gets the ?-free "What about you:" phrasing, while X, Bluesky,
>   and Mastodon keep "What about you?". A comment notes the LinkedIn quirk so nobody
>   reintroduces it.
> 
>   Worth verifying with a real LinkedIn account before merging, since this behavior is
>   entirely on LinkedIn's side and undocumented — the diagnosis rests on the prod-vs-PR
>   comparison in Daria's screenshots. If you'd rather not have per-network phrasing, the
>   alternative is reverting to "What about you:" everywhere.

```diff
diff --git a/src/components/island/BingoCard.svelte b/src/components/island/BingoCard.svelte
index 085796e6..1688e91c 100644
--- a/src/components/island/BingoCard.svelte
+++ b/src/components/island/BingoCard.svelte
@@ -48,18 +48,20 @@
     } catch {}
   }
 
-  function buildShareText() {
+  function buildShareText(forLinkedIn = false) {
     const attended = editions.filter((_, i) => checked[i]).map(e => `${e.year} ${e.city}`);
     const count = attended.length;
     const base = `I've attended ${count} EuroPython conference${count !== 1 ? 's' : ''}! 🐍`;
     const body = count > 0 ? `${base}\n\n${attended.join(' · ')}` : base;
-    return `${body}\n\nWhat about you? ${BINGO_PAGE_URL}\n\n#EuroPython #Python`;
+    // LinkedIn drops everything after a "?" in the prefilled text, even when encoded
+    const prompt = forLinkedIn ? 'What about you:' : 'What about you?';
+    return `${body}\n\n${prompt} ${BINGO_PAGE_URL}\n\n#EuroPython #Python`;
   }
 
   const BINGO_PAGE_URL = 'https://ep2026.europython.eu/bingo';
 
   function shareLinkedIn() {
-    const text = buildShareText();
+    const text = buildShareText(true);
     window.open(
       `https://www.linkedin.com/feed/?shareActive=true&text=${encodeURIComponent(text)}`,
       '_blank', 'noopener,noreferrer'

```

But we can just change "What about you?" back to "What about you:" for a quick fix.

dariagrudzien:
> But we can just change "What about you?" back to "What about you:" for a quick fix.

Sure, that makes sense. Let's make sure to keep 2 newlines before the hashtags, so the url pops out.

I'm heading for the EP and some family time before that, will be mostly offline from now on.

src/components/island/BingoCard.svelte
```diff
@@ -53,10 +53,10 @@
     const count = attended.length;
     const base = `I've attended ${count} EuroPython conference${count !== 1 ? 's' : ''}! 🐍`;
     const body = count > 0 ? `${base}\n\n${attended.join(' · ')}` : base;
-    return `${body}\n\nWhat about you: ${BINGO_PAGE_URL}`;
+    return `${body}\n\nWhat about you? ${BINGO_PAGE_URL}\n\n#EuroPython #Python`;
```
hugovk:
```suggestion
    return `${body}\n\nWhat about you: ${BINGO_PAGE_URL}\n\n#EuroPython #Python`;
```

hugovk:
> > But we can just change "What about you?" back to "What about you:" for a quick fix.
> 
> Sure, that makes sense. Let's make sure to keep 2 newlines before the hashtags, so the url pops out.

Done, back to colon, and with 2 newlines before the hashtags.

> I'm heading for the EP and some family time before that, will be mostly offline from now on.

Have a good break before EP!

Perhaps someone else can approve/merge.

hypha (approved):
Tested on LinkedIn and looks good now.

nikoshell:
> @hugovk I don't really know why, but this change breaks LinkedIn sharing by removing the URL, by clipping everything after the last question including the URL:
> 
> this PR image
> 
> prod image

We should check our **main page** OpenGraph details to be sure that there is no issue when we share it on socials.

nikoshell:
Not related to this PR.
#1732

PR #1714, opened 2026-07-01
Added Beginners' Day content

t-redactyl:
I've added in a page for the Django Girls and Humble Data events on the first sprints day.


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2026-beginners-day-page.ep-preview.click|
|last update|2026-07-03T22:54:29.886Z|

package.json
nikoshell, line 1: Do we need this change?

t-redactyl: You're right, I didn't mean to commit this one. No, we don't need to change package.json.

t-redactyl:
@nikoshell I restored package.json to the version on main (ep2026). Let me know if I did it right.

nikoshell (approved):

PR #1707, opened 2026-06-30
New schedule view with advanced search

nikoshell:
Schedule view with filters and multiple views (timeline, grid, list)

Unpublished pages are marked as red in the main navigation just for this PR.
They will be visible as red only in dev instance.

PR #1735, opened 2026-07-03
Fix timline overlaps

nikoshell:


nikoshell:
This PR is fixing this case
image

europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2026-fix-schedule-time.ep-preview.click|
|last update|2026-07-05T13:48:51.376Z|

hypha:
Thanks for working on this! It looks better than before. But right now, the time for Friday 14:30 is for the glass room, which is for community events, instead of the talk rooms. To me, it makes more sense to: 
- put glass room to after S4
- show talk room's start times (they will also be more nicely aligned!)
- in each glass room's sessions, we can maybe put a small xx:xx - yy:yy (or just the start time), so people still know when it starts, but the main timeline reflects the main talk schedule. 
WDYT?

nikoshell:
> Thanks for working on this! It looks better than before. But right now, the time for Friday 14:30 is for the glass room, which is for community events, instead of the talk rooms. To me, it makes more sense to:
> 
>     * put glass room to after S4
> 
>     * show talk room's start times (they will also be more nicely aligned!)
> 
>     * in each glass room's sessions, we can maybe put a small xx:xx - yy:yy (or just the start time), so people still know when it starts, but the main timeline reflects the main talk schedule.
>       WDYT?

We noticed the issues with rooms you mention and try for improve it with progamme api 
One of the solution which we working on is multiple schedule views with one column time line and view with grid.
Here is draft #1707 but there are also issue related to schedule which have higher priority.

nikoshell:
@hypha is there anything else to do in scope of this issue ? 
this PR solves only overlaping on timeline

hypha:
Hi @nikoshell the overlapping issue is indeed solved. I just think it should show the start time of the main tracks (S1-S4) on the timeline, instead of the community event. But if that is better addressed in #1707, then totally fine to merge this one!

nikoshell:
We try to fix bugs quickly and merge quickly.

Feature or design requests can go with regular workflow.

hypha (approved):
If I understand correctly, we do improvement in new PRs. Approving as this solves the overlapping issue.

nikoshell:
I will write it down and prepare PR with that

PR #1739, opened 2026-07-04
Schedule: sort rooms with S-number first

hugovk:
Wednesday looks like:
image

But when you go to Thursday or Friday, suddenly "Glass Room" appears and moves the keynotes to the right:

image

Instead, sort rooms without an S-number ("Glass room", "Fishbowl") last:

image

---

This also means Monday/Tuesday change from summits then tutorials:

image

To tutorials then summits:

image

But I think that's fine.

### DIFF
--- a/src/components/schedule/day.astro
+++ b/src/components/schedule/day.astro
@@ -37,7 +37,8 @@ const dayName = day.id;
 //const day = getEntry("days", dayName);
 
 const getRoomSortKey = (room: string) => {
-  const match = room.match(/\(S(\d+)([A-Z]?)\)/i);
+  // rooms without an S-number ("Glass room", "Fishbowl") sort last
+  const match = room.match(/^S(\d+)([A-Z]?)$/i);
 
   if (!match) {
     return `99-${room}`;
--- a/src/components/schedule/day.astro
+++ b/src/components/schedule/day.astro
@@ -36,21 +36,31 @@ const { day } = Astro.props;
 const dayName = day.id;
 //const day = getEntry("days", dayName);
 
+// Predefined display order; rooms not listed here sort last, alphabetically.
+const ROOM_ORDER = [
+  "S1",
+  "S2",
+  "S3A",
+  "S3B",
+  "S4",
+  "S4A",
+  "S4B",
+  "Glass room",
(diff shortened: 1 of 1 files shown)


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://sort-rooms.ep-preview.click|
|last update|2026-07-05T13:19:21.654Z|

artcz:
Thanks for the patch!

We were discussing the problem with Marcin a few days ago, and in general I think it's better if we go with a predefined order of rooms – either coming directly from pretalx, or from programmeapi, or  hardcoded in the the website. Some of the new things on the talk pages will require a predefined room order.

A stable sorting algorithm would be also be good enough, but explicit order might be simpler and easier to understand and maintain.

cc @nikoshell

egeakman:
I think it'd be better to do the predefined order here in the website.

nikoshell:
> I think it'd be better to do the predefined order here in the website.

Yep, we want do it on the website.

hugovk:
I can update this PR.

Here's the list of rooms:

- 2.017/2.018
- Exhibit Hall
- Fishbowl
- Glass room
- S1
- S2
- S3A
- S3B
- S4
- S4A
- S4B

What order would you like?

hypha:
> What order would you like?

I think this

Tutorial days: 
- S3A
- S3B
- S4A
- S4B
- 2.017/2.018
- Fishbowl

Conf days: 
- S1
- S2
- S3A
- S3B
- S4
- Glass room
- Exhibit Hall (for posters) floats on lunch break so not sure if the order matters for the purpose of the schedule display.

hugovk:
Updated, looks like:

image

image

image

image

image

hypha:
If rooms are not listed in the order, or if we do not give a pre-defined order, what would the behaviour be? To me, what makes sense in terms of sorting priorities: 
- pre-defined order
- pretalx order
- alphabetical order

The current new order in preview help me breathe better, so thank you!

hugovk:
This PR sorts by the predefined order, and for rooms not listed there, they're put in alphabetical order at the end.

hypha (approved):

PR #1751, opened 2026-07-07
Add wifi ssid and password

egeakman:
so the mobile app can fetch the latest creds from here. We should update this file when the creds change.


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://wifi.ep-preview.click|
|last update|2026-07-07T00:26:56.478Z|

PR #1757, opened 2026-07-08
Use h1 for page titles to improve a11y, SEO, site search

hugovk:
Some pages uses h2 for titles, when the top-level should be h1.

I spotted some search results had "undefined" headers:

image

This is because Pagefind uses the h1 for the title by default.

Using a single h1 header also helps accessibility -- screen readers benefit from being able to semantically find the main page header -- and SEO.

`src/components/ui/Title.astro` is used for some pages that should have an h1 header, but also for in places which should remain as h2, such as homepage sections such as `sections/krakow.astro`. So it's h2 by default, and is explicitly set to h1 where needed.

### DIFF
--- a/src/components/ui/Title.astro
+++ b/src/components/ui/Title.astro
@@ -4,6 +4,7 @@ export interface Props {
   subtitle?: string;
   color?: string; // title text color, default #fff
   anchorColor?: string; // anchor link color, default matches color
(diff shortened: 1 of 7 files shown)


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://h1.ep-preview.click|
|last update|2026-07-08T10:57:15.906Z|

nikoshell (approved):

PR #1767, opened 2026-07-09
Fix posters 

nikoshell:


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2026-fix-posters.ep-preview.click|
|last update|2026-07-10T19:00:34.924Z|

hypha:
Much better, thank you! Is it possible or a good idea to write the start time of the posters? On Thursday, I expect the posters to show before the lunch. But on the timeline, it is after. So I wonder if people might miss 10 min of the session. And I am not sure if PyLadies Lunch should be there.

nikoshell:
That was exactly the request,
posters always shown during lunchtime
and events like PyLadies and Community Organizers below posters at lunchtime  

image

hypha:
>  and events like PyLadies and Community Organizers below posters at lunchtime

Sorry I was not aware of the PyLadies Lunch. Thank you for the clarification. 

>  posters always shown during lunchtime

Yes, but I am wondering if it can show the start time of the posters on each poster card, so that people will know they are not after 13:35 as indicated in the timeline. maybe something like this: 
screenshot of the poster sessions showing start-finish time, instead of durations

gbdlin:
LGTM

@hypha I think the desired start hour is the same as lunch, but it is wrongly set in pretalx.

gbdlin (approved):

nikoshell:
We plant to create one more PR and change style of side events during lunch time.

PR #1795, opened 2026-07-13
sprint project proposel

radioactivityy:
🖼️ Preview available 🖼️ : https://ep-website--1795.org.readthedocs.build/

### DIFF
--- a/src/content/sprints/_sprints_template.md
+++ b/src/content/sprints/_sprints_template.md
@@ -1,24 +1,24 @@
 ---
 title: "Real time dynamic pricing AI system in Python"
-numberOfPeople: "5" 
-pythonLevel: "Intermediate" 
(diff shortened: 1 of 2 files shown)


nikoshell (changes requested):
Please don't change template file src/content/sprints/_sprints_template.md

rodrigogiraoserrao:
@radioactivityy sorry, the file `_sprints_template.md` is still different from what it was before your PR.

Also, for your sprint proposal, you'll want to remove `draft: true` so it's published.

Finally, sprint proposals are _typically_ to collaborate on existing open source projects. It looks like you may be looking for people to start a new project with you, which is really cool! It may be a good idea to be clearer about that in the project description so that your intentions are clearer :D

PR #1808, opened 2026-07-15
Add NumPy sprint

ngoldbaum:
Ping @seberg in case you're interested in being a contact as well.

🖼️ Preview available 🖼️ : https://ep-website--1808.org.readthedocs.build/


nikoshell (approved):

seberg:
Thanks for the ping.  Could add me, but no priority we can see on spot (let's see maybe I'll add a CuPy slot or steer that way depending on interest).

PR #1822, opened 2026-07-16
Fix pyladies room change

nikoshell:


nikoshell:
Session is already visible in Fishbowl, only issue is that we have 6 columns now.

europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2026-fix-schedule.ep-preview.click|
|last update|2026-07-16T12:33:54.841Z|

PR #1845, opened 2026-08-08
add component DateGate conditionally render content based on build-time date

nikoshell:


europython-society:
# Preview available
|Key|Value|
|---|---|
|url|https://ep2026-scope.ep-preview.click|
|last update|2026-08-08T09:25:08.711Z|

src/components/Header.astro
```diff
@@ -111,12 +112,24 @@ const activeMenus = NAV_MENUS.map((menu) => ({
...
         )
     }
-    <li class="nav-item nav-mobile-cta">
-      <a class="nav-link-btn" href="/jobs">Jobs</a>
-    </li>
-    <li class="nav-item nav-mobile-cta">
-      <a class="nav-link-btn" href="/tickets">Register Now</a>
-    </li>
+    <DateGate before="2026-07-20">
+      <Fragment slot="before">
```
hypha, line 116: I’m confused here. so when only `before="2026-07-20"` is provided, the `beforeDate` selects `during` up to 20 Jul and `after` afterward, right? If so, it never selects `before`, so this `slot="before"` fragment  doesn't reach / render? Then isn't it just a dup of  `during` fragment?

Also, could we give an example to the the before-only case in `DateGate` to help with the confusedpoke?

src/components/Header.astro
```diff
@@ -160,8 +173,21 @@ const activeMenus = NAV_MENUS.map((menu) => ({
...
+    <Fragment slot="before">
+      <a href="/jobs" class="nav-cta-desktop">Jobs</a>
+      <a href="/tickets" class="nav-cta">Register Now</a>
+    </Fragment>
+    <Fragment slot="during">
+      <a href="/jobs" class="nav-cta-desktop">Jobs</a>
+      <a href="/tickets" class="nav-cta">Register Now</a>
+    </Fragment>
+    <Fragment slot="after">
+      <span class="nav-cta" style="opacity:0.5;cursor:default"
```
hypha, line 186: opacity: 0.5 here fades both the txt and bg into the navbar, so in lightmode, the "see you next year" is so faint that I barely think you mean it. 😢 Maybe remove the opacity and give this post-event label its own txt and bg colour, or something else cleverer, to increase the contrast and sincerity of wanting to see us next year?

src/components/ui/DateGate.astro
```diff
@@ -0,0 +1,79 @@
...
+}
+
+const { after, before } = Astro.props;
+
+// --- Parse dates ---
+// Date-only strings (no "T") get interpreted as UTC midnight by new Date().
+// DateTime strings with offset (e.g. "...T00:00:00+02:00") parse correctly.
+const parseDate = (s: string): Date => {
+  const iso = s.includes("T") ? s : `${s}T00:00:00.000Z`;
+  return new Date(iso);
```
hypha, line 54: nit: is it a good idea for parseDate to check if it produced a valid date and throw a build error if not? Like if i have a typo, e.g. before="2026-13-20", that's an invalid date,  I think in the before-only case, comparing now with that invalid date returns false, so we select `during` and will continue to show the gated content instead of reporting the typo?

hypha (approved):
Some minor non-blocking comments and sorry for the delay.

